PR #6, opened 2020-03-17
Catch and log ToolsErrors in main entry point

rwalton-arm:
### Description

Subclass `click.Group` so we can catch and log `ToolsErrors`

### Test Coverage

### DIFF
--- a/mbed_tools/cli.py
+++ b/mbed_tools/cli.py
@@ -1,15 +1,51 @@
 """Integration point with all sub-packages."""
+import logging
+
 import click
+
+from mbed_tools_lib.logging import log_exception, set_log_level
+from mbed_tools_lib.exceptions import ToolsError
+
 from mbed_devices.mbed_tools import cli as mbed_devices_cli
 
 
 CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
+LOGGER = logging.getLogger(__name__)
 
 
-@click.group(context_settings=CONTEXT_SETTINGS)
-def cli() -> None:
-    """Main entry point for the cli."""
-    pass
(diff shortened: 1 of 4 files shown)


rwalton-arm:
This is dependent on an `mbed-tools-lib` release as I changed the signature of `log_exception`.

mbed_tools/cli.py
```diff
@@ -1,15 +1,51 @@
...
+        Args:
+            context: The current click context.
+        """
+        try:
+            super().invoke(context)
+        except ToolsError as err:
+            traceback = context.params["traceback"]
+            if LOGGER.level != logging.DEBUG and not traceback:
+                err = (
+                    f"{err} Increase the verbosity level to `-vvv` to see debug information."
```
muchzill4: Would you be able to provide an example error output in the cli? I wonder if this line is readable as it currently stands. Does this rely on every `{err}` ending with a dot? 🤔

rwalton-arm: Example output
```
ERROR: Failed to find a connected device. Increase the verbosity level to `-vvv` to see debug information. Use the `--traceback` argument to see a full stack trace.
```

Yes it relies on people writing sentences that end with a full stop. Otherwise we get `{err}.. Increase the...` sometimes, which looks silly.

muchzill4: Awesome, thanks. Do you think separating that output into two lines would improve things or not? I.e. f"{err}\nIncrease..". Wonder if that slight change would make this more sentence proof. :)

rwalton-arm: Yeah, sounds good.

rwalton-arm: 436857a

PR #7, opened 2020-03-17
✨ Read configuration from child packages

muchzill4:
### Description

Output configuration variables exposed by dependencies when calling `mbed-tools config` sub command.

Note, build will fail as it requires releases of both `mbed-devices` and `mbed-targets`.

### Test Coverage


mbed_tools/_internal/config_cli.py
```diff
@@ -0,0 +1,46 @@
...
+from mbed_devices.mbed_tools import config_variables as mbed_devices_config_variables
+
+
+@click.command()
+def cli() -> None:
+    """Prints information about configuration variables of all child mbed packages."""
+    click.echo(_build_output(_gather_configuration_variables()))
+
+
+def _gather_configuration_variables() -> List[pdoc.Variable]:
```
rwalton-arm: Any reason we don't just pass `mbed_devices_config_variables` directly to `_build_output`? I don't think this single line function buys us much when we only have a single caller?

muchzill4: You're right, that's silly.

muchzill4: https://github.com/ARMmbed/mbed-tools/pull/7/commits/828bd2539eb02dfd4dcdda15fe7cc79c2bddb7aa

setup.py
```diff
@@ -34,7 +34,7 @@
     description="Command line interface for Mbed OS.",
     keywords="Arm Mbed OS MbedOS cli command line tools",
     include_package_data=True,
-    install_requires=["python-dotenv", "Click==7.0", "mbed-devices", "mbed_tools_lib"],
+    install_requires=["python-dotenv", "Click==7.0", "mbed-devices", "pdoc3"],
```
rwalton-arm: We do need `mbed-tools-lib` for the `log_exception` function.

muchzill4: https://github.com/ARMmbed/mbed-tools/pull/7/commits/470c3143173b2c6332bff8fa0b4f20c73ced164c

PR #30, opened 2020-07-22
Merge tools sub repos into single repo

rwalton-arm:
### Description

A multiple repo / package hierarchy with many different versions of dependencies vastly increases the maintenance burden.

This PR makes the following changes:
* Merge all tools packages into a single repo so we can simplify dependency management and release processes.
* Remove the subpackage clis and move all the cli entry points into a "cli" package folder at the top level.
* Move to a `src` directory structure; this is so we can be sure our unit tests are representative of the installed package and not the state of the current working directory (pytest automatically prepends the `cwd` to `sys.path` so we could've been missing packaging bugs with the previous tree structure).

### Test Coverage


setup.cfg
```diff
@@ -38,14 +37,15 @@ ignore =
     H306,
     # W503:  line break before binary operator (this is no longer PEP8 compliant)
     W503,
+    # E203: whitespace before ':' (this is not PEP8 compliant)
+    E203,
```
Patater, line 41: Why ignore E203?

Why ignore E203 in a "Move to src dir layout" commit?

setup.py
```diff
@@ -52,15 +52,16 @@
...
+    packages=find_packages(where="src"),
     python_requires=">=3.6,<4",
     url=f"https://github.com/ARMmbed/{PROJECT_SLUG}",
     version=__version__,
     entry_points={
         "console_scripts": [
-            f"mbedtools={SOURCE_DIR}.cli:cli",
-            f"mbed-tools={SOURCE_DIR}.cli:cli",
-            f"mbed_tools={SOURCE_DIR}.cli:cli",
+            "mbedtools=mbed_tools.cli.cli:cli",
```
Patater: .cli.cli:cli looks funny. Is that correct? so many .cli?

setup.py
```diff
@@ -46,7 +46,10 @@
         "mbed-devices~=1.0",
         "mbed-project~=2.0",
         "pdoc3",
```
Patater, line 45: "merge mbed_tools_lib package" commit title should start with a capital letter.

Patater: Which commit were these copied from? Note the source repo and commit in the commit message.

src/mbed_tools/project/_internal/git_utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+@dataclass
+class GitReference:
+    """Git reference for a remote repository.
+
+    Attributes:
+        repo_url: URL of the git repository.
+        ref: The reference commit sha, tag or branch.
+    """
+
```
Patater, line 23: Does black not want to remove this newline? I see in `clone()` below we don't have a newline after the docstring.

src/mbed_tools/targets/_internal/data/targets_metadata.json
```diff
@@ -0,0 +1,121 @@
+{
```
Patater, line 1: Should we remove `targets_metadata.json` in a separate PR?

setup.py
```diff
@@ -51,6 +51,10 @@
         "tabulate",
         "dataclasses; python_version<'3.7'",
         "requests>=2.20",
+        "pywin32; platform_system=='Windows'",
+        "psutil; platform_system=='Linux'",
```
Patater, line 52: Probably pre-existing, but how about on Mac?

src/mbed_tools/devices/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+#
+"""API to detect any Mbed OS devices connected to the host computer.
+
+It is expected that this package will be used by developers of Mbed OS tooling rather than by users of Mbed OS.
+This package uses the https://github.com/ARMmbed/mbed-targets interface to identify valid Mbed Enabled Devices.
+Please see the documentation for mbed-targets for information on configuration options.
+
+For the command line interface to the API see the package https://github.com/ARMmbed/mbed-tools
+"""
+from mbed_tools.devices.devices import get_connected_devices
```
Patater, line 13: Does this need to be "devices.devices" or is there a nicer or flatter name we can use?

Pipfile
```diff
@@ -20,6 +20,11 @@ mbed-tools-ci-scripts = "*"
 pre-commit = "*"
 requests-mock = "*"
 factory_boy = "*"
+# Additional requirements for ./ci_scripts
+boto3 = "*"
+jinja2 = "*"
```
Patater, line 25: Shouldn't this have been added to the Pipfile in the same commit that added this requirement in setup.py (i.e. in "merge mbed-build package")?

setup.py
```diff
@@ -65,9 +65,9 @@
     version=__version__,
     entry_points={
         "console_scripts": [
-            "mbedtools=mbed_tools.cli.cli:cli",
-            "mbed-tools=mbed_tools.cli.cli:cli",
-            "mbed_tools=mbed_tools.cli.cli:cli",
+            "mbedtools=mbed_tools.cli.main:cli",
```
Patater, line 68: Ah, great. You give a better name here.

Patater: Do you think we should do the "anything `mbed-` becomes a subcommand of mbedtools" thing? For example, when we want to allow others to plug in new commands. Do you think the commit "Remove sub package CLIs" makes this easier or harder to do in the future? I suspect we'd want each sub command to expose their own cli and be installed as e.g. "mbed-configure" for "mbedtools configure".

src/mbed_tools/cli/main.py
```diff
@@ -37,22 +37,12 @@ def invoke(self, context: click.Context) -> None:
 
 
 def print_version(context: click.Context, param: Union[click.Option, click.Parameter], value: bool) -> Any:
-    """A click callback which prints the versions of all Mbed Python packages."""
+    """A click callback that prints the versions of mbed-tools."""
```
Patater, line 40: Should this be "version" instead of "versions"? Does mbed-tools have multiple versions?

azure-pipelines/build-release.yml
```diff
@@ -54,7 +54,7 @@ stages:
             Windows_Py_3_8:
               python.version: '3.8'
               vmImageName: windows-latest
-              uploadCoverage: "false"
+              uploadCoverage: "true"
```
Patater, line 57: If this is what broke the CI result overview, we should reconsider making this change until that issue is resolved.

CONTRIBUTING.md
```diff
@@ -63,39 +63,37 @@ The `pre-commit` tool allows developers to easily install git hook scripts which
 must pass in our CI before a PR is merged. Using commit hooks ensures you can't commit code which violates our style
 and maintainability requirements.
 
-To install the commit hooks for the repository, run `pipenv install --dev` then `pipenv shell`, in the `pipenv shell`
-type `pre-commit install`, the checks will now run automatically every time you try to `git commit` to the repository.
+To install the commit hooks for the repository, run:
+```bash
+pipenv install --dev
```
Patater, line 68: Do we want to add `$` to make it clear these are shell prompts?

CONTRIBUTING.md
```diff
@@ -63,39 +63,37 @@ The `pre-commit` tool allows developers to easily install git hook scripts which
...
-the changes:
+We aim to compose and review patchsets, not individual patches.
+When merging a pull request, we preserve the commit history. Ensure your commit history is clean and understandable.
 
-- `<emoji> <Issue-Number> <Change Summary> (#<Pull Request Number>)`
+We recommend the following commit structure in the following order:
 
-An emoji is used to highlight what has occurred in the change. Commonly used emojis can be seen below, but for a full
-list please see [Gitmoji](https://gitmoji.carloscuesta.me/):
+* Whitespace only changes.
```
Patater, line 86: Use a numbered list to emphasize that ordering is important.

Patater, line 86: Remove `.` from the end of list items

setup.cfg
rwalton-arm: We ignore `E203` as it's not compatible with `black` or `Pep8`. Granted that probably should've been in the "merge mbed-build" commit (as that's the repo it came from), or a separate commit.

Patater: OK, it looked like this came out of the blue. Fine with this going into mbed-build merge commit.

src/mbed_tools/project/_internal/git_utils.py
rwalton-arm: No, that's expected. According to `Pep257`, in class definitions a newline is required, in function definitions it's not. https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings - `black` formats this way.

src/mbed_tools/targets/_internal/data/targets_metadata.json
rwalton-arm: IIRC `mbed-targets` now `.targets` requires it. I'll investigate that and raise another PR if we can remove it.

src/mbed_tools/project/_internal/git_utils.py
Patater: OK, thanks

setup.py
rwalton-arm: I change it later when moving the cli's to top level.

src/mbed_tools/targets/_internal/data/targets_metadata.json
Patater: OK

Patater: I made a sticky note for it

setup.py
rwalton-arm: these were pre-existing here, this commit removes `mbed_tools_lib` (below) as this is the commit that merges it.

src/mbed_tools/cli/main.py
rwalton-arm: It was printing the versions for all installed `mbed-` packages. I missed that, will change it.

azure-pipelines/build-release.yml
rwalton-arm: Unfortunately then the coverage is reported as below 50%, as `mbed-devices` contained os specific code.

rwalton-arm: So some tests were skipped on different platforms.

CONTRIBUTING.md
rwalton-arm: Hmm, that's debatable, but I'll go with it.

setup.py
rwalton-arm: These aren't required on mac, it's using bundled macos command line tools to check for the USB devices and doesn't rely on libraries AFAICT.

src/mbed_tools/devices/__init__.py
rwalton-arm: `devices.get_devices`? We're importing to the package level here in `__init__.py`, so calling clients will import `devices.get_connected_devices` but I agree it looks noisy.

Pipfile
rwalton-arm: The requirement is added in "merge mbed-targets ci_scripts", not `mbed-build`.

setup.py
rwalton-arm: Moving our internal commands shouldn't make it any harder to do that. If we use `https://pypi.org/project/click-plugins/` we only need to add the plugin commands to the main group. We should do it, but not as part of this merge.

setup.py
Patater: Hmm, we'll need something for BSD probably

Patater: I made a sticky note for it.

rwalton-arm: Yeah, we'll probably need to implement BSD specific code IIUC.

CONTRIBUTING.md
Patater: At least be consistent.

The style I follow is only to add `.` at the end of bullet items if the items are complete sentences.

setup.py
rwalton-arm: I'm not sure we want the packages that rely on core APIs (which are `targets`, `lib` and `project`) to be plugin modules (`configure` and `list_devices` both rely on these). If we want `configure` et al to actually be plugins, we shouldn't merge them into the same repo, and we should go back to the old structure of independent repos for every subpackage. Standalone packages (possibly `pyocd` and any others) could be converted to plugins and would be installed separately. Thoughts?

loverdeg:
Has the previous team been consulted on this?
I will always be in support of digging in to avoid a monolith.

Resources are probably thin though, so I get why this PR exists.

DEVELOPMENT.md
```diff
@@ -2,15 +2,21 @@
 
 For development and testing purposes, it is essential to use a virtual environment. It is recommended that `pipenv >= 2018.11.26` is used.
 
-## Setup Pipenv
+## Setup Python and Pipenv
 
-To start developing, install pip and pipenv on your system. Note the latter is done at user level to keep the system installation of python clean which is important on a Mac (at least):
+`mbed-toools` is compatible with Python 3.6 or later.
```
Patater, line 7:
```diff
- `mbed-toools`
+ `mbed-tools`
```

Patater:
> Has the previous team been consulted on this?
> I will always be in support of digging in to avoid a monolith.

Unfortunately, we already had a monolith, given that we often needed to coordinate changes across multiple repos (landing a PR in each without breaking tests along the way has been a challenge.)

This change should reduce the maintenance burden by putting things that need to change together into one repository. This will make working on the tools more efficient.

To avoid additional coupling between submodules, we will be adding more automated tests to find and error on submodule interdependencies. We'll also work to eliminate the coupling that is already present. After which, we can decide if the benefits of breaking into separate repos again outweigh the costs of doing so (could be).

src/mbed_tools/devices/__init__.py
Patater: OK

Pipfile
Patater: OK

setup.py
Patater: I agree it makes sense to have some core commands that aren't implemented as plugins. Let's consider which commands make most sense as plugins after we 1) reduce the coupling between our submodules further, and 2) have the plugin mechanism determined.

azure-pipelines/build-release.yml
Patater: OK, good not to hide this test coverage gap.

PR #31, opened 2020-07-23
ADR for removing mbed config header

rwalton-arm:
### Description

Add an ADR for removing mbed_config.h

### Test Coverage

- [x]  Additional tests are not required for this change (e.g. documentation update).

### DIFF
--- a/news/20200723142705.doc
+++ b/news/20200723142705.doc
@@ -1,0 +1,1 @@
+Add ADR for removing mbed_config.h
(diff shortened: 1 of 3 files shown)


news/20200723142705.doc
```diff
@@ -0,0 +1 @@
+Add ADR for removing mbed_config.h
```
Patater, line 1: ADRs probably don't need NEWS items, as they are developer-facing rather than user-facing. NEWS is for users; developers can use `git log`.

rwalton-arm: We should tweak the CI so it doesn't try to fail the PR when there's no news fragment.

PR #34, opened 2020-07-27
Do not install editable version of package in ci

rwalton-arm:
### Description

The CI pipeline was using editable installs, so the tests failed to catch any packaging issues.

We now install the package to `site-packages` in CI, so we're testing what we actually deploy.

### Test Coverage

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -3,6 +3,9 @@ name = "pypi"
 url = "https://pypi.org/simple"
 verify_ssl = true
 
+[packages]
+mbed-tools = {editable = true, path = "."}
+
 [dev-packages]
 black = "*"
 coverage = "*"
@@ -15,7 +18,6 @@ mypy = ">=0.500"
 pytest = "*"
 pytest-cov = "*"
 wheel = "*"
-mbed-tools = {editable = true, path = "."}
 mbed-tools-ci-scripts = "*"
 pre-commit = "*"
 requests-mock = "*"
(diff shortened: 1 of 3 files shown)


Pipfile
```diff
@@ -3,6 +3,9 @@ name = "pypi"
 url = "https://pypi.org/simple"
 verify_ssl = true
 
+[packages]
+mbed-tools = {editable = true, path = "."}
```
Patater, line 7: Why do we want mbed-tools to be editable at all?

rwalton-arm: Some people find it useful during development as any changes are immediately picked up when you run the program. We definitely don't want it to be editable in the CI.

I don't generally find it useful to run an editable install, as there are some subtle cases where changes aren't picked up so you have to reinstall the package anyway. I didn't want to just go and remove it, and it didn't make sense to install `pytest` et al as `packages`. If you run `pipenv install -e .` it puts this in `packages` so I thought that would be the least surprising. However, I don't particularly like `pipenv` or editable installs so I went with what I thought was principle of least surprise for people who enjoy such things.

rwalton-arm:
> I'm a bit confused by this. I'd expect the final "packages" verison not to be "editable", and the "dev-packages" version to be editable if needed for convenience or otherwise. Could CI use a non-editable "packages" version?

We can't do that as described, as the editable package version installation is then coupled with required development dependencies used in CI, putting us back where we started. We need to separate out the package install from the dev/ci dependencies installation, which is how I've done it here. The CI now only installs dev dependencies, and we use `pip install .` to install the package, which installs a non editable wheel to `site-packages`.

My proposal for future is to remove the `Pipfile` and move to `tox` for development/CI dependencies, as it has far more flexible workflow. It also means we aren't tied to any particular CI system, as all the CI will do is run `tox` with different sets of command line arguments for analysis, testing, doc building, etc. This will make it easier to move to Travis or another CI provider in the future if needed.

PR #38, opened 2020-07-29
Ensure correct exit code is returned when ToolsError is raised

rwalton-arm:
### Description

When `ToolsError` exceptions were raised the tools returned an exit code of 0.

Return 1 when `ToolsError` is raised.

### Test Coverage

### DIFF
--- a/src/mbed_tools/lib/logging.py
+++ b/src/mbed_tools/lib/logging.py
@@ -35,6 +35,7 @@ class MbedToolsHandler:
         """Keep track of the logger to use and whether or not a traceback should be generated."""
         self._logger = logger
         self._traceback = traceback
+        self.exit_code = 0
 
     def __enter__(self) -> "MbedToolsHandler":
         """Return the Context Manager."""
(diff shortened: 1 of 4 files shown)


src/mbed_tools/lib/logging.py
```diff
@@ -51,6 +52,7 @@ def __exit__(
             error_msg = _exception_message(cast(BaseException, exc_value), logging.root.level, self._traceback)
             self._logger.error(error_msg, exc_info=self._traceback)
             # Do not propagate exceptions derived from ToolsError
```
Patater, line 54: Is this comment still accurate now that we are propagating an exit code? I'm not sure what this comment is referring to nor why we wouldn't want to propagate errors or exceptions.

rwalton-arm: It's still accurate, but not informative. This code doesn't propagate `ToolsErrors` and subclasses, as the thinking was just to log those errors as they are "expected errors" (whatever that means). It was a requirement that the user didn't see a traceback in the case of a `ToolsError` unless they asked for it explicitly on the command line  (i.e through a `--traceback` argument).

 If a non `ToolsError` is raised, that is supposed to mean the error hadn't been handled by the code and was a bug, so let the exception propagate up to the interpreter. The way I've done things in the past is to catch all `Exception`s at the top level and only show a traceback if the user asks for it on the command line. I don't entirely understand the `ToolsError` and non `ToolsError` reasoning.

rwalton-arm: We should fix up the unintuitive `ToolsError` vs non `ToolsError` handling. I propose just catching everything inheriting from `Exception` at the top level and removing this context manager entirely. We should do that in a separate PR, though.

PR #41, opened 2020-08-04
Move pytest config to pytest.ini, add norecursedirs

rwalton-arm:
### Description

The "build" folder is specified in the default pytest norecursedirs
setting, meaning the tests for the build module were not executed.

Pytest also recommend not using setup.cfg for configuration, as the
setup.cfg parser can cause hard to diagnose problems. So move the pytest
config to pytest.ini as that is the most stable config format.

This commit also deletes irrelevant failing tests and fixes a failing
test in test_cmake_file.py, which were not being flagged as tests were
not executed.

### Test Coverage


tests/build/__init__.py
```diff
@@ -2,4 +2,3 @@
 # Copyright (C) 2020 Arm Mbed. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-"""Tests."""
```
Patater, line 5: But this was so informative! I don't see how we'll live without this.

PR #42, opened 2020-08-04
Make the configure command use the mbed program root directory as output

rwalton-arm:
### Description

The configure command wasn't using the MbedProgram object to find the root directory of the project. Meaning if the user was in a program subdirectory and didn't specify an output path, the config file would be created in the wrong place and the build would fail.

While I'm here, I've tried to remove unnecessary instantiations of the `MbedProgram` and `Target` objects and decouple them from the `build` module. Previously we were creating multiple `MbedProgram` and `Target` objects in different functions. I've pulled the creation of these objects up to the top level command handler. This code could still be improved further but I think this is enough of an improvement for this PR.

### Test Coverage

### DIFF
--- a/src/mbed_tools/cli/configure.py
+++ b/src/mbed_tools/cli/configure.py
@@ -7,8 +7,6 @@ import pathlib
 
 import click
 
-from typing import Any
-
 from mbed_tools.project import MbedProgram
 from mbed_tools.targets import get_target_by_name
 from mbed_tools.build._internal.cmake_file import generate_mbed_config_cmake_file
@@ -19,16 +17,6 @@ from mbed_tools.build._internal.write_files import write_file
 
 @click.command(
     help="Generate an Mbed OS config CMake file and write it to a .mbedbuild folder in the program directory."
-)
-@click.option(
-    "-o",
-    "--output-directory",
-    type=click.Path(),
-    default=".",
-    help=(
-        "Destination for the generated .mbedbuild/mbed_config.cmake file containing configuration parameters to build "
-        "Mbed OS. The default is the current working directory."
-    ),
 )
 @click.option(
     "-t",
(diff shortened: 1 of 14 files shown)


src/mbed_tools/cli/configure.py
```diff
@@ -66,6 +53,6 @@ def configure(output_directory: Any, toolchain: str, mbed_target: str, program_p
     target_build_attributes = get_target_by_name(mbed_target, program.mbed_os.targets_json_file)
     config = assemble_config(Source.from_target(target_build_attributes), program.root, program.files.app_config_file)
     cmake_file_contents = generate_mbed_config_cmake_file(mbed_target, target_build_attributes, config, toolchain)
-    output_directory = pathlib.Path(output_directory, ".mbedbuild")
+    output_directory = program.root / ".mbedbuild"
```
Patater, line 56: I'm unfamiliar with this syntax. You can really divide by a string and append to the path? If not, probably we should use some sort of path joining function.

tests/cli/test_configure.py
```diff
@@ -0,0 +1,42 @@
...
+        assemble_config.assert_called_once_with(
+            source.from_target.return_value,
+            mbed_program.from_existing.return_value.root,
+            mbed_program.from_existing.return_value.files.app_config_file,
+        )
+        gen_config_cmake.assert_called_once_with(
+            "K64F", get_target_by_name.return_value, assemble_config.return_value, "GCC_ARM"
+        )
+        write_file.assert_called_once_with(
+            mbed_program.from_existing.return_value.root / ".mbedbuild",
```
Patater, line 39: Unfamiliar syntax here, too.

src/mbed_tools/cli/configure.py
rwalton-arm: `pathlib.Path` overloads the `/` operator to perform the same function as `join`.

Patater: Thanks for explaining. I figured it'd be some operator overloading magic given the unit tests pass and this code is included in the tests.

PR #45, opened 2020-08-06
project: Remove dependency on .mbed file

rwalton-arm:
### Description

If a user manually git clones an Mbed project, and it doesn't contain a
.mbed file, they would have to create it manually. This is not good UX.
Instead check for the presence of an mbed-os.lib file, as this file is
required in all Mbed programs.

### Test Coverage

### DIFF
--- a/src/mbed_tools/project/mbed_program.py
+++ b/src/mbed_tools/project/mbed_program.py
@@ -14,7 +14,7 @@ from mbed_tools.project._internal import git_utils
 from mbed_tools.project._internal.project_data import (
     MbedProgramFiles,
     MbedOS,
-    PROGRAM_ROOT_FILE_NAME,
+    MBED_OS_REFERENCE_FILE_NAME,
     MBED_OS_DIR_NAME,
 )
 from mbed_tools.project._internal.libraries import LibraryReferences, MbedLibReference
@@ -42,7 +42,7 @@ class MbedProgram:
         """
         self.repo = repo
         self.files = program_files
-        self.root = self.files.mbed_file.parent
+        self.root = self.files.mbed_os_ref.parent
         self.mbed_os = mbed_os
         self.lib_references = LibraryReferences(root=self.root, ignore_paths=[self.mbed_os.root])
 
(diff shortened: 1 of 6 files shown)


src/mbed_tools/project/mbed_program.py
```diff
@@ -191,33 +191,32 @@ def _tree_contains_program(path: Path) -> bool:
...
         potential_root = potential_root.parent
 
-    logger.debug("No .mbed file found.")
+    logger.debug("No mbed-os.lib file found.")
     raise ProgramNotFound(
-        f"No program found from {cwd.resolve()} to {cwd.resolve().anchor}. Please set the cwd to a program directory "
-        "containing a .mbed file. You can also set your cwd to a program subdirectory if there is a .mbed file at the "
-        "root of your program's directory tree. If your program does not contain a .mbed file, please create an empty "
-        ".mbed file at the root of the program directory tree before performing any other operations."
+        f"No program found from {cwd.resolve()} to {cwd.resolve().anchor}. Please set the path to a program directory "
```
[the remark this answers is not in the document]
rwalton-arm: The user can also pass in a path, which is why I changed it from "current working directory". I'll come up with something less ambiguous.

tests/project/_internal/test_mbed_program.py
```diff
@@ -90,7 +90,6 @@ def test_from_existing_raises_if_path_is_not_a_program(self, fs):
     def test_from_existing_raises_if_program_files_missing(self, fs):
         fs_root = pathlib.Path(fs, "foo")
         fs_root.mkdir()
-        (fs_root / ".mbed").touch()
```
jainvikas8, line 93: should we `(fs_root /  "mbed-os.lib").touch()` do this here?

rwalton-arm: I'm not sure what you mean. We've removed all references to `.mbed` so we don't need to test it here.

jainvikas8: If you see `test_from_new_local_dir_raises_if_path_is_existing_program`  was replaced `(program_root / ".mbed").touch()` ... unless I got `test_from_existing_raises_if_program_files_missing` test wrong?

rwalton-arm: Sorry, I see what you mean now... So before we checked for `.mbed` and `mbed-os.lib`, and this test was just checking `mbed-os.lib`, so we had to touch `.mbed`, so it didn't raise in the wrong place. Now `.mbed` isn't checked, so we raise in the wrong place (in _find_program_root, instead of MbedProgram.from_existing). I'll rework it a bit so it's more obvious. Good catch.

rwalton-arm: I've removed some duplicated error handling in `MbedProgramFiles` and deleted this test.

src/mbed_tools/project/mbed_program.py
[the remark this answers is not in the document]
rwalton-arm: Changed "path" to "directory".

PR #46, opened 2020-08-06
Generate project startup files from "init" command

rwalton-arm:
### Description

Generate a CMakeLists.txt, main.cpp and .gitignore from the "init" command

Depends on: https://github.com/ARMmbed/mbed-tools/pull/45

### Test Coverage

### DIFF
--- a/src/mbed_tools/project/_internal/templates/gitignore.tmpl
+++ b/src/mbed_tools/project/_internal/templates/gitignore.tmpl
@@ -1,0 +1,2 @@
+.mbedbuild
+cmake_build/
--- a/src/mbed_tools/project/_internal/templates/main.tmpl
+++ b/src/mbed_tools/project/_internal/templates/main.tmpl
@@ -1,0 +1,13 @@
+/* mbed Microcontroller Library
+ * Copyright (c) {{year}} ARM Limited
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "mbed.h"
+
+
+int main()
+{
+    printf("Hello, Mbed!\n");
+    return 0;
+}
--- a/src/mbed_tools/project/mbed_program.py
+++ b/src/mbed_tools/project/mbed_program.py
@@ -32,15 +32,13 @@ class MbedProgram:
         * A collection of references to external libraries, defined in .lib files located in the program source tree
     """
 
-    def __init__(self, repo: git_utils.git.Repo, program_files: MbedProgramFiles, mbed_os: MbedOS) -> None:
+    def __init__(self, program_files: MbedProgramFiles, mbed_os: MbedOS) -> None:
         """Initialise the program attributes.
 
         Args:
-            repo: The program's git repository.
             program_files: Object holding paths to a set of files that define an Mbed program.
             mbed_os: An instance of `MbedOS` holding paths to locations in the local copy of the Mbed OS source.
         """
-        self.repo = repo
         self.files = program_files
         self.root = self.files.mbed_os_ref.parent
         self.mbed_os = mbed_os
@@ -63,7 +61,7 @@ class MbedProgram:
                 "to an empty directory."
             )
         logger.info(f"Cloning Mbed program from URL '{url}'.")
-        repo = git_utils.clone(url, dst_path)
+        git_utils.clone(url, dst_path)
 
         program_files = MbedProgramFiles.from_existing(dst_path)
         if not program_files.mbed_os_ref.exists():
@@ -79,7 +77,7 @@ class MbedProgram:
         except ValueError as mbed_err:
             raise MbedOSNotFound(f"{mbed_err}")
 
-        return cls(repo, program_files, mbed_os)
+        return cls(program_files, mbed_os)
 
     @classmethod
     def from_new(cls, dir_path: Path) -> "MbedProgram":
@@ -103,9 +101,8 @@ class MbedProgram:
         dir_path.mkdir(exist_ok=True)
         program_files = MbedProgramFiles.from_new(dir_path)
         logger.info(f"Creating git repository for the Mbed program '{dir_path}'")
-        repo = git_utils.init(dir_path)
         mbed_os = MbedOS.from_new(dir_path / MBED_OS_DIR_NAME)
-        return cls(repo, program_files, mbed_os)
+        return cls(program_files, mbed_os)
 
     @classmethod
     def from_existing(cls, dir_path: Path, check_mbed_os: bool = True) -> "MbedProgram":
@@ -123,7 +120,6 @@ class MbedProgram:
         logger.info(f"Found existing Mbed program at path '{program_root}'")
         program = MbedProgramFiles.from_existing(program_root)
 
-        repo = git_utils.get_repo(program_root)
         try:
             mbed_os = MbedOS.from_existing(program_root / MBED_OS_DIR_NAME, check_mbed_os)
         except ValueError as mbed_os_err:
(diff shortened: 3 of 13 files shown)


codecov.yml
```diff
@@ -2,6 +2,8 @@ codecov:
   require_ci_to_pass: yes
   notify:
     after_n_builds: 3
+  ignore:
```
jainvikas8, line 5: I think it's better to remove this file, because in future if the file get new code added to it, then it may be missed from coverage or no need to ignore it

rwalton-arm: We don't add code in `__init__.py` as it's bad practice. It should only be used for exporting modules to a package's public interface. We can't remove the file as the packaging system requires it.

jainvikas8: Ok, there are some many `__init__.py` which can be part of this commit., to improve Code coverage

jainvikas8: Example : `src\mbed_tools\lib\__init__.py` will this be included to?

rwalton-arm: `**/*` is a recursive glob in an attempt to ignore all of them. Although looking at the report it's not ignoring `templates/__init__.py` for some reason, I'll see if that's my fault or codecov's.

tests/project/_internal/test_project_data.py
```diff
@@ -45,6 +71,32 @@ def test_from_existing_finds_existing_program_data(self, fs):
...
+    def test_from_existing_calls_render_cmakelists_template_with_program_name_if_file_nonexistent(self, fs):
+        with mock.patch("mbed_tools.project._internal.project_data.render_jinja_template") as render_cmake:
+            root = pathlib.Path(fs, "foo")
+            make_mbed_program_files(root)
+            render_cmake.return_value = "foo"
+
+            MbedProgramFiles.from_existing(root)
+
+            render_cmake.assert_called_with(
+                CMAKELISTS_TEMPLATE_NAME, {"program_name": root.stem, "year": str(datetime.datetime.now().year)}
```
Patater, line 87: May need to mock `datetime` here as well.

src/mbed_tools/project/_internal/templates/gitignore.tmpl
```diff
@@ -0,0 +1,2 @@
+.mbedbuild
+build/
```
Patater, line 2: Should we use `cmake_build` for the folder name so as not to clash with the old build system's BUILD folder on case-insensitive file systems like what Windows often employs?

codecov.yml
```diff
@@ -2,6 +2,8 @@ codecov:
...
+    - "**/*__init__.py"
```
Patater, line 6: Should this be `**/__init__.py`? Do we intend to ignore files named like `mcbaz__init__.py`?

src/mbed_tools/project/mbed_program.py
```diff
@@ -63,7 +61,7 @@ def from_url(cls, url: str, dst_path: Path, check_mbed_os: bool = True) -> "Mbed
                 "to an empty directory."
             )
         logger.info(f"Cloning Mbed program from URL '{url}'.")
-        repo = git_utils.clone(url, dst_path)
+        git_utils.clone(url, dst_path)
```
Patater, line 64: Do we have any tests covering where this clone fails? Maybe the `feature-cmake` branch or whatever reference in the `mbed-os.lib` file is broken. I know we have a test `test_raises_when_clone_fails()` that checks an exception is thrown from `clone()`, but I'm not sure we have anything testing how we handle invalid references in mocked mbed-os.lib files.

codecov.yml
rwalton-arm: I don't think this even works as it stands. I also tried `**/__init__.py` and it still incorrectly flagged up the `templates/__init__.py`. I'll see if I can fix it.

src/mbed_tools/project/_internal/templates/gitignore.tmpl
jainvikas8: Normally follow : BUILD folder (caps)

src/mbed_tools/project/mbed_program.py
rwalton-arm: At the moment we just propagate any error thrown from git. We don't do any extra checking of the lib references. We probably should. I'll see if I can add something.

rwalton-arm: If we just let git deal with invalid references we get the following error message:

```
$ mbedtools init ../../programs/mytestprog
Creating a new Mbed program at path '../../programs/mytestprog'.
Downloading mbed-os and adding it to the project.
ERROR: Failed to check out revision 'featurhakhkahkah'. Error from VCS:
  stderr: 'error: pathspec 'featurhakhkahkah' did not match any file(s) known to git'
```

This seems OK to me, do you think we need to do anything more here? Maybe we could add which `.lib` file is corrupted to the error message?

codecov.yml
rwalton-arm: Reverted this change as it wasn't having the desired effect.

jainvikas8: Might be worth adding an individual entry for each file. I don't think there are a lot of them, which may also improve code readability to figure out "where to start"

src/mbed_tools/project/mbed_program.py
Patater: That error looks fine. Should we have a test to ensure that the error from git is propagated as expected?

codecov.yml
Patater: I see you've removed this for now.

tests/project/_internal/test_render_templates.py
```diff
@@ -0,0 +1,38 @@
...
+# Copyright (C) 2020 Arm Mbed. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from unittest import TestCase, mock
+
+from mbed_tools.project._internal.render_templates import (
+    render_cmakelists_template,
+    render_main_cpp_template,
```
Patater, line 11: Should we also test render_gitignore_template() here?

tests/project/_internal/test_project_data.py
```diff
@@ -5,29 +5,50 @@
...
-    def test_from_new_returns_valid_program(self, fs):
+    def test_from_new_returns_valid_program_file_set(self, fs):
         root = pathlib.Path(fs, "foo")
         root.mkdir()
 
         program = MbedProgramFiles.from_new(root)
 
         self.assertTrue(program.app_config_file.exists())
+        self.assertTrue(program.mbed_os_ref.exists())
+        self.assertTrue(program.cmakelists_file.exists())
```
Patater, line 37: Should we also check for a freshly created gitignore?

src/mbed_tools/project/mbed_program.py
rwalton-arm: We do have a test to ensure we're propagating the error here: https://github.com/ARMmbed/mbed-tools/blob/8800c3d07a7e31bb1c477eebe88fb68887b873cd/tests/project/_internal/test_git_utils.py#L64

codecov.yml
rwalton-arm: I don't think it's worth it for "code readability". If I was looking for "where to start" I'd rather see a code style guide. It's kinda unexpected to study the codecov config file for stuff like that. We should have a style guide, though. I will add a task for it to the backlog.

rwalton-arm: Added a task for code style guide https://jira.arm.com/browse/IOTBTOOL-957

src/mbed_tools/project/_internal/templates/main.tmpl
```diff
@@ -0,0 +1,13 @@
+/* mbed Microcontroller Library
+ * Copyright (c) {{year}} ARM Limited
```
jainvikas8, line 2: Will it support the year format `2020 - 2021` or `2020,2021` etc

rwalton-arm: Don't think it needs to. This is just the year of the creation of the file.

PR #47, opened 2020-08-11
Fix config processing label collection

rwalton-arm:
### Description

The config processing code wasn't collecting and target, component or feature labels from the config system.

### Test Coverage


src/mbed_tools/targets/_internal/target_attributes.py
```diff
@@ -80,12 +80,13 @@ def _read_json_file(path_to_file: pathlib.Path) -> Any:
         FileNotFoundError: path provided does not lead to a valid json file
     """
     try:
-        return json.loads(path_to_file.read_text())
+        # mypy doesn't recognise that json.loads always returns a dictionary
+        return cast(dict, json.loads(path_to_file.read_text()))
     except JSONDecodeError as json_err:
```
Patater, line 88: Are there potential exceptions from the cast we should handle here or should we just let the exception bubble up?

tests/build/_internal/test_cmake_file.py
```diff
@@ -11,14 +11,15 @@
...
-        target.labels = ["foo"]
-        target.features = ["bar"]
-        target.components = ["baz"]
-        target.macros = ["macbaz"]
-        target.device_has = ["stuff"]
-        target.core = ["core"]
-        target.supported_form_factors = ["arduino"]
+        target = dict()
+        target["labels"] = ["foo"]
+        target["extra_labels"] = ["foo"]
```
Patater, line 16: Shall we try "morefoo" or "extrafoo" for extra uniqueness?

src/mbed_tools/targets/get_target.py
```diff
@@ -26,15 +25,16 @@ def get_target_by_name(name: str, path_to_targets_json: pathlib.Path) -> Target:
     Raises:
         TargetError: an error has occurred while fetching target
     """
-    return Target.from_targets_json(name, path_to_targets_json)
+    try:
+        return target_attributes.get_target_attributes(path_to_targets_json, name)
+    except (FileNotFoundError, target_attributes.TargetAttributesError) as e:
+        raise TargetError(e) from e
```
Patater, line 31: Should we add a test to pretend to cause a file not found and check we get TargetError?

Patater:
With the dropping of `-D`, do we also need CMake changes or is this rather fixing the tools to catch up with CMake changes that already happened?

src/mbed_tools/targets/_internal/target_attributes.py
rwalton-arm: At runtime `cast` is a no-op, it only executes any code when type checking

rwalton-arm:
> With the dropping of `-D`, do we also need CMake changes or is this rather fixing the tools to catch up with CMake changes that already happened?

We don't need CMake changes `target_compile_definitions` ignores any `-D` https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

src/mbed_tools/targets/get_target.py
rwalton-arm: Yeah, this is also why codecov is failing and why the PR is still marked "WIP".

rwalton-arm: Added tests to check `TargetError` is raised when we hit `FileNotFound` or `TargetAttributesError`.

tests/build/_internal/config/test_assemble_build_config.py
```diff
@@ -96,11 +102,60 @@ def test_assembles_config_using_all_relevant_files(self):
...
+            created_mbed_app_file = create_files(directory, [mbed_app_file])[0]
+
+            _ = _assemble_config_from_sources_and_lib_files(
+                target, find_files("mbed_lib.json", Path(directory)), created_mbed_app_file
+            )
+
+            self.assertEqual(target["features"], {"RED", "BLUE"})
+            self.assertEqual(target["components"], {"LEG"})
+            self.assertEqual(target["extra_labels"], {"EXTRA_HOT"})
+            self.assertEqual(target["labels"], {"A", "PICKLE"})
```
Patater, line 161: Should we check `target["macros"]` or `target["config"]` here as well?

rwalton-arm: `target["config"]` is handled by the `Config` object and is checked in another test. I have added a check for `target["macros"]` here.

PR #51, opened 2020-08-14
ci: Add travis config file

urutva:
### Description
Enable Travis-CI to validate latest changes on supported mbed-os
examples.

### Test Coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,103 @@
+# Copyright (c) 2020 Arm Limited. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the License); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+language: sh
+os: linux
+dist: xenial
+
+cache:
+  pip: true
+  ccache: true
+  # It looks like ccache for arm-none-eabi is not yet supported by Travis.
+  # Therefore manually adding ccache directory to cache
+  directories:
+    - ${HOME}/.ccache
+
+addons:
+  apt:
+    packages:
+    - ninja-build
+
+matrix:
+  include:
+    - &mbed-tools-test
+      stage: "Mbed-Tools"
+      name: "Test blinky example - develop (K64F)"
+      env: EXAMPLE_NAME=mbed-os-example-blinky TARGET_NAME=K64F PROFILE=develop CACHE_NAME=develop-K64F
+      language: python
+      python: 3.8
+      install:
+        # Setup ccache
+        - ccache -o compiler_check=content
+        - ccache -M 1G
+        - pushd /usr/lib/ccache
+        - sudo ln -s ../../bin/ccache arm-none-eabi-gcc
+        - sudo ln -s ../../bin/ccache arm-none-eabi-g++
+        - export PATH="$PATH:/usr/lib/ccache"
+        - popd
+        # Install arm-none-eabi-
+        - pushd /home/travis/build && mkdir gcc-arm && cd gcc-arm
+        - curl -L0 "https://developer.arm.com/-/media/Files/downloads/gnu-rm/9-2020q2/gcc-arm-none-eabi-9-2020-q2-update-x86_64-linux.tar.bz2?revision=05382cca-1721-44e1-ae19-1e7c3dc96118&la=en&hash=D7C9D18FCA2DD9F894FD9F3C3DC9228498FA281A" --output gcc-arm-none-eabi-9-2020-q2-update.tar.bz2
+        - tar xf gcc-arm-none-eabi-9-2020-q2-update.tar.bz2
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,94 @@
...
+        - tar xf gcc-arm-none-eabi-9-2020-q2-update.tar.bz2
+        - export PATH="${PWD}/gcc-arm-none-eabi-9-2020-q2-update/bin:$PATH"
+        - popd
+        - arm-none-eabi-gcc --version
+        - pip install --upgrade cmake
+        # Install the changes added in the pull request
+        - pip install git+https://github.com/ARMmbed/mbed-tools.git@refs/pull/${TRAVIS_PULL_REQUEST}/head
+        - mbedtools --version
+        # We use manual clone, with depth and single branch = the fastest
+        # Because of this, we need to create .mbed file as the current tools require it
```
Patater, line 53: No, no `.mbed` file needed. The comment is wrong.

.travis.yml
```diff
@@ -0,0 +1,94 @@
...
+      # Install arm-none-eabi-gcc
+        - pushd /home/travis/build && mkdir arm-gcc && cd arm-gcc
+        - curl -L0 "https://developer.arm.com/-/media/Files/downloads/gnu-rm/9-2020q2/gcc-arm-none-eabi-9-2020-q2-update-x86_64-linux.tar.bz2?revision=05382cca-1721-44e1-ae19-1e7c3dc96118&la=en&hash=D7C9D18FCA2DD9F894FD9F3C3DC9228498FA281A" --output gcc-arm-none-eabi-9-2020-q2-update.tar.bz2
+        - tar xf gcc-arm-none-eabi-9-2020-q2-update.tar.bz2
+        - export PATH="${PWD}/gcc-arm-none-eabi-9-2020-q2-update/bin:$PATH"
+        - popd
+        - arm-none-eabi-gcc --version
+        - pip install --upgrade cmake
+        # Install the changes added in the pull request
+        - pip install git+https://github.com/ARMmbed/mbed-tools.git@refs/pull/${TRAVIS_PULL_REQUEST}/head
```
rwalton-arm, line 50: Would `cd ${TRAVIS_BUILD_DIR} && pip install .` not work instead of fetching again from github?

urutva: In order to do that, the contents of the PR needs to be fetched. Directly installing from Github seems simpler.

rwalton-arm: Travis doesn't automatically fetch and check out your PR? Seems strange to me. I need to read some more about how Travis works then.

urutva: It turns out that Travis will fetch the contents of the PR. So changing the install command to `pip install .`.

.travis.yml
```diff
@@ -0,0 +1,94 @@
...
+      python: 3.8
+      install:
+      # Install arm-none-eabi-gcc
+        - pushd /home/travis/build && mkdir arm-gcc && cd arm-gcc
+        - curl -L0 "https://developer.arm.com/-/media/Files/downloads/gnu-rm/9-2020q2/gcc-arm-none-eabi-9-2020-q2-update-x86_64-linux.tar.bz2?revision=05382cca-1721-44e1-ae19-1e7c3dc96118&la=en&hash=D7C9D18FCA2DD9F894FD9F3C3DC9228498FA281A" --output gcc-arm-none-eabi-9-2020-q2-update.tar.bz2
+        - tar xf gcc-arm-none-eabi-9-2020-q2-update.tar.bz2
+        - export PATH="${PWD}/gcc-arm-none-eabi-9-2020-q2-update/bin:$PATH"
+        - popd
+        - arm-none-eabi-gcc --version
+        - pip install --upgrade cmake
```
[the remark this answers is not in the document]
urutva: `--upgrade` ensures that latest version is installed. If the `cmake` is already installed then `pip install cmake` will not do anything.

.travis.yml
```diff
@@ -0,0 +1,101 @@
...
+      name: "CMake blinky example - develop (NRF52840_DK)"
+      env: NAME=cmake_test EXAMPLE_NAME=mbed-os-example-blinky TARGET_NAME=NRF52840_DK PROFILE=develop
+
+    - <<: *cmake-build-test
+      name: "CMake blinky example - release (NRF52840_DK)"
+      env: NAME=cmake_test EXAMPLE_NAME=mbed-os-example-blinky TARGET_NAME=NRF52840_DK PROFILE=release
+
+    - <<: *cmake-build-test
+      name: "CMake blinky example - debug (NRF52840_DK)"
+      env: NAME=cmake_test EXAMPLE_NAME=mbed-os-example-blinky TARGET_NAME=NRF52840_DK PROFILE=debug
```
jainvikas8, line 101: Do we need to mention all the configs separately?  Can we have a loop which reads different targets and profiles instead?

Patater: I prefer this to a loop. cmake_test is parameterized and each of these is just calling it. This is flat and easy to understand. When we have hundreds of targets, this will be very long: it may make sense to have a tool to help auto-generate this file. When we are rapidly adding targets, to decrease the maintenance burden of updating each of our travis.yml (in mbed-os, blinky, and also in tools as here), it may make sense to make the job smarter to detect which targets are supported and build each profile for them. For now, this is good enough, IMO.

rwalton-arm: FWIW I prefer this flat, declarative structure too. CI scripts shouldn't have anything clever in them ;)

jainvikas8: We could use a loop to build all profiles, as they are pretty much a standard, if possible. Thanks.

urutva: There doesn't seems to be a looping construct in Travis configuration https://docs.travis-ci.com/user/speeding-up-the-build to parallelise builds. As @Patater recommended, in future we can think about auto-generating `.travis.yml` file. But for time being, I'll leave this as it is.

Patater (approved):

PR #59, opened 2020-08-20
Fix generate-target-database azure pipeline

rwalton-arm:
### Description

Requires an editable install in CI so the board database path is found by the githelpers package.

### Test Coverage


ci_scripts/sync_board_database.py
```diff
@@ -172,6 +165,13 @@ def main(args: argparse.Namespace) -> int:
         else:
             news_file_text = "Offline board database created."
 
+        pr_info = PullRequestInfo(
+            repo="ARMMbed/mbed-tools",
+            head_branch=args.head_branch,
+            base_branch=args.base_branch,
+            subject=args.pr_subject,
+            body=news_file_text,
+        )
```
Patater, line 174: Could you commend why using news file text as the PR description is useful? What motivates the change?

azure-pipelines/generate-target-database.yaml
```diff
@@ -30,7 +30,13 @@ jobs:
         inputs:
           versionSpec: '3.7'
 
-      - template: steps/install-development-dependencies.yml
```
Patater, line 33: Can we delete the file `steps/install-development-dependencies.yml`?

Patater (changes requested):

ci_scripts/sync_board_database.py
rwalton-arm: Will do.

azure-pipelines/generate-target-database.yaml
rwalton-arm: Not yet, it's still used in the `build-and-release` pipeline. When we move to `tox` we won't need it anymore.

PR #74, opened 2020-09-24
ci: Compare binaries of old and new tools

urutva:
### Description

Compare binaries of old and new tools

### Test Coverage

- [x ]  This change is covered by existing or additional automated tests.


urutva:
This approach is not useful as the binaries generated with old and new tools doesn't match. However, testing the binary generated with cmake works as expected on the target.

PR #77, opened 2020-09-29
Add build command porcelain

rwalton-arm:
### Description

Add build command porcelain.

### Test Coverage

### DIFF
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -1,0 +1,63 @@
+#
+# Copyright (C) 2020 Arm Mbed. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Command to build an Mbed project using CMake."""
+import os
+import pathlib
+
+import click
+
+from mbed_tools.build import build_project, generate_build_system, generate_config
+from mbed_tools.project import MbedProgram
+
+
+@click.command(name="build", help="Build an Mbed project.")
+@click.option(
+    "-t",
+    "--toolchain",
+    type=click.Choice(["ARM", "GCC_ARM"], case_sensitive=False),
+    help="The toolchain you are using to build your app.",
+)
+@click.option("-m", "--mbed-target", help="A build target for an Mbed-enabled device, e.g. K64F.")
+@click.option("-b", "--build-type", default="develop", help="The build type (release, develop or debug).")
+@click.option("-c", "--clean", is_flag=True, default=False, help="Perform a 'clean' build.")
+@click.option(
+    "-p",
+    "--program-path",
+    default=os.getcwd(),
+    help="Path to local Mbed program. By default it is the current working directory.",
+)
+def build(program_path: str, build_type: str, toolchain: str = "", mbed_target: str = "", clean: bool = False) -> None:
+    """Configure and build an Mbed project using CMake and Ninja.
+
+    If the project has already been configured and contains '.mbedbuild/mbed_config.cmake', this command will skip the
+    Mbed configuration step and invoke CMake.
+
+    If the CMake configuration step has already been run previously (i.e a CMake build tree exists), then just try to
+    build the project immediately using Ninja.
+
+    Args:
+       program_path: Path to the Mbed project.
+       build_type: The Mbed build profile (debug, develop or release).
+       toolchain: The toolchain to use for the build.
+       mbed_target: The name of the Mbed target to build for.
+       clean: Force regeneration of config and build system before building.
+    """
+    program = MbedProgram.from_existing(pathlib.Path(program_path))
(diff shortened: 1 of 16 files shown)


tests/cli/test_configure.py
```diff
@@ -12,30 +10,9 @@
...
-    @mock.patch("mbed_tools.cli.configure.get_target_by_name")
-    @mock.patch("mbed_tools.cli.configure.generate_mbed_config_cmake_file")
-    @mock.patch("mbed_tools.cli.configure.assemble_config")
-    @mock.patch("mbed_tools.cli.configure.write_file")
-    def test_collaborators_called_with_corrrect_arguments(
-        self, write_file, assemble_config, gen_config_cmake, get_target_by_name, mbed_program
-    ):
-        CliRunner().invoke(configure, ["-m", "k64f", "-t", "GCC_ARM"])
+    def test_generate_config_called_with_correct_arguments(self, program, generate_config):
+        CliRunner().invoke(configure, ["-m", "k64f", "-t", "gcc_arm"])
```
Patater, line 16: Does case matter here? Should we keep using GCC_ARM as before, and also start using K64F?

src/mbed_tools/cli/build.py
```diff
@@ -0,0 +1,63 @@
...
+    default=os.getcwd(),
+    help="Path to local Mbed program. By default is the current working directory.",
+)
+def build(program_path: str, build_type: str, toolchain: str = "", mbed_target: str = "", clean: bool = False) -> None:
+    """Configure and build an Mbed project using CMake and Ninja.
+
+    If the project has already been configured and contains '.mbedbuild/mbed_config.cmake', this command will skip the
+    Mbed configuration step and invoke CMake.
+
+    If the CMake configuration step has already been ran previously (i.e a CMake build tree exists), then just try to
```
Patater, line 37:
```diff
- been ran
+ been run
```

src/mbed_tools/cli/build.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@click.command(name="build", help="Build an Mbed project.")
+@click.option(
+    "-t",
+    "--toolchain",
+    type=click.Choice(["ARM", "GCC_ARM"], case_sensitive=False),
+    help="The toolchain you are using to build your app.",
+)
+@click.option("-m", "--mbed-target", help="A build target for an Mbed-enabled device, eg. K64F")
```
Patater, line 22:
```diff
- eg.
+ e.g.
```

src/mbed_tools/cli/build.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Args:
+       program_path: Path to the Mbed project.
+       build_type: The Mbed build profile (debug, develop or release).
+       toolchain: The toolchain to use for the build.
+       mbed_target: The name of the Mbed target to build for.
+       clean: Force regeneration of config and build system before building.
+    """
+    program = MbedProgram.from_existing(pathlib.Path(program_path))
+    mbed_config_file = program.root / ".mbedbuild" / "mbed_config.cmake"
```
Patater, line 48: Doesn't this duplicate the knowledge of where the mbed config file lives? Should we use a path querying function for the module that owns there the config goes instead? It's harder to change where the config file goes or what its name is with the current duplication.

tests/cli/test_build.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@contextlib.contextmanager
+def mock_project_directory(mbed_config_exists=False, build_tree_exists=False):
+    with TemporaryDirectory() as tmp_dir:
+        root = pathlib.Path(tmp_dir, "test-project")
+        root.mkdir()
+        build_tree = root / "cmake_build"
+        mbed_config_dir = root / ".mbedbuild"
+        mbed_config_file = mbed_config_dir / "mbed_config.cmake"
```
Patater, line 24: Would tests that use this mock stop working if we renamed the config file? Should we query the path from a module that owns the path here instead? It looks like this mock is "by chance" matching what the other module uses, but that could change if the module changes.

Patater (changes requested):
Could we update `.travis.yml` to build using the new porcelain? That'd improve our existing integration tests there.

src/mbed_tools/cli/build.py
rwalton-arm: I had it in my mind to add the path to the config file as an `MbedProgram` attribute, but I forgot to do that. Good catch.

tests/cli/test_configure.py
rwalton-arm: Case does matter. The `targets` API needs the target name in uppercase. The toolchain is just passed through to mbed_config.cmake, I believe, and CMake also expects uppercase.

Patater (approved):
LGTM

PR #78, opened 2020-09-29
configure: Don't emit _PRESENT macros

Patater:
### Description

Do not put macros like MBED_CONF_FILESYSTEM_PRESENT or
MBED_CONF_EVENTS_PRESENT into the applicaiton's generated
mbed_config.cmake file. We will depend on Mbed OS CMake files for each
of these components to define the _PRESENT macro in a
target_compile_definitions() block for the e.g. filesystem or events
CMake targets.

This should be merged immediately before [the Mbed OS companion PR](https://github.com/ARMmbed/mbed-os/pull/13732) is ready. Then, we should merge this PR, do a tools release, merge Mbed OS example PRs, and merge the Mbed OS PR. Without Mbed OS having updated CMake files, these tool changes will break some builds.

### Test Coverage

- [X]  Additional tests are not required for this change (e.g. documentation update).

### DIFF
--- a/news/20200929160957.misc
+++ b/news/20200929160957.misc
@@ -1,0 +1,1 @@
+Don't emit _PRESENT macros. Depend on Mbed OS to define these from CMake.
--- a/src/mbed_tools/build/_internal/config/config.py
+++ b/src/mbed_tools/build/_internal/config/config.py
@@ -104,6 +104,14 @@ class Config:
         config = Config()
         for source in sources:
             for key, value in source.config.items():
+                # If the config item is about a certain component or feature
+                # being present, avoid adding it to the mbed_config.cmake
+                # configuration file. Instead, applications should depend on
+                # the feature or component with target_link_libraries() and the
+                # component's CMake flle (in the Mbed OS repo) will create
+                # any necessary macros or definitions.
+                if key.endswith(".present"):
+                    continue
                 _create_config_option(config, key, value, source)
             for key, value in source.overrides.items():
                 if key in IGNORED_OVERRIDE_KEYS_IN_SOURCE:


src/mbed_tools/build/_internal/cmake_file.py
```diff
@@ -47,7 +47,7 @@ def _render_mbed_config_cmake_template(
     """
     env = jinja2.Environment(loader=jinja2.PackageLoader("mbed_tools.build", str(TEMPLATES_DIRECTORY)),)
     template = env.get_template(TEMPLATE_NAME)
-    options = list(config.options.values())
+    options = [x for x in config.options.values() if not x.key.endswith(".present")]
```
rwalton-arm, line 50: Would it not be better to apply this filtering where we create the config "options" here https://github.com/ARMmbed/mbed-tools/blob/495311ca64b90ef068dbc2d9a01e7fb927666615/src/mbed_tools/build/_internal/config/config.py#L106? If not could we at least have a comment as to why we filter this and maybe a `TODO` to remove this when we refactor the mbed config?

Patater: I'm not sure what refactoring of mbed config you have in mind here. Perhaps we should raise a GitHub issue for it?

rwalton-arm (approved):
LGTM

src/mbed_tools/build/_internal/cmake_file.py
rwalton-arm: I meant refactoring of the mbed configuration system to remove these macros from the mbed_lib.json files.

hugueskamba (approved):

PR #81, opened 2020-10-05
Use tox instead of pipenv

jainvikas8:
### Description

Replace pipenv usage with Tox

- Linting
- Analyze and test runs
- Build docs
- Generate Licensing (SPDX)
- Checknews

### Test Coverage

- [ x]  This change is covered by existing or additional automated tests.


azure-pipelines/build-release.yml
```diff
@@ -26,50 +26,83 @@ stages:
         strategy:
           maxParallel: 10
           matrix:
+            Assert_news:
+              python.version: '3.7'
```
Patater, line 30: Why not 3.8?

news/202010061300.misc
```diff
@@ -0,0 +1 @@
+Replace pipenv with tox
```
Patater, line 1: I think we've done more than just replace pipenv with tox (generate docs, generate spdx). Should we have other news files, too?

azure-pipelines/build-release.yml
```diff
@@ -26,50 +26,77 @@ stages:
         strategy:
           maxParallel: 10
           matrix:
+            Linting_Py_3_6:
```
Patater, line 35: Commit "ci: Use tox instead of pipenv for analyse and test" doesn't do anything with pipenv. Should it be called instead "ci: Use tox to run linting and pytest"?

azure-pipelines/build-release.yml
```diff
@@ -150,11 +150,12 @@ stages:
             inputs:
               versionSpec: '3.7'
 
-          - template: steps/install-development-dependencies.yml
-
-          - template: steps/generate-documentation.yml
+          - script: |
```
Patater, line 138: I don't see any pipenv removal in commit "ci: Use tox instead of pipenv for docs build". Should we name the commit "ci: Use tox for docs build"?

azure-pipelines/generate-target-database.yaml
```diff
@@ -32,10 +32,7 @@ jobs:
 
       - script: |
           python -m pip install --upgrade pip wheel setuptools
-          pip install pipenv
-          python -m pipenv lock --dev-only -r --pre > dev-requirements.txt
-          pip install -r dev-requirements.txt
-          pip install pytest-azurepipelines
+          pip install -r requirements-test.txt
```
Patater, line 35: Unlike the other commits, "ci: remove pipenv from generate-target-database.yaml" does indeed remove use of pipenv: this is good. Unlike the other commits, the first word after the change area is lowercase: we should use "ci: Remove pipenv from generate-target-database.yaml" to make this PR self-consistent.

azure-pipelines/build-release.yml
```diff
@@ -174,7 +174,8 @@ stages:
             inputs:
               versionSpec: '3.7'
 
-          - template: steps/install-development-dependencies.yml
+          - script: |
```
Patater, line 162: Commit "ci: Use tox instead of pipenv for licensing" doesn't appear to remove use of pipenv. It should be renamed "ci: Use tox to generate the license report"

azure-pipelines/build-release.yml
```diff
@@ -129,9 +129,8 @@ stages:
             inputs:
               versionSpec: '3.7'
 
-          - template: steps/install-development-dependencies.yml
-
           - script: |
+              python -m pip install --upgrade tox
```
Patater: Commit "ci: Use tox instead of pipenv for checknews" doesn't remove pipenv. We should rename the commit to "ci: Use tox to run the news check"

Patater (commented):
Should we add a commit at the end that removes `azure-pipelines/steps/install-development-dependencies.yml` now that it isn't used anymore?

azure-pipelines/build-release.yml
jainvikas8: Previously was using python 3.7, so just retaining was done before, I don't think this would matter if 3.8 or 3.7

news/202010061300.misc
jainvikas8: I have made the news file for each commit, explaining further changes.

jainvikas8:
> Should we add a commit at the end that removes `azure-pipelines/steps/install-development-dependencies.yml` now that it isn't used anymore?

It's been used Beta and Production release, will be raised as new PR.

jainvikas8:
Forced pushed to address comments from @Patater .

azure-pipelines/build-release.yml
Patater: I was thinking if we were adding new stuff, why wouldn't we want to use the latest python. If it was 3.7 before, this is fine.

Patater (approved):
LGTM

PR #89, opened 2020-10-16
Add supported_application_profiles target config

rajkan01:
### Description

- Added the target config supported_application_profiles support
- Updated the existing test to validate supported_application_profiles target config
- Added the news file

### Test Coverage

- [x]  Manual testing has been performed (and evidence provided) as automated testing was not feasible.

### DIFF
--- a/news/20201023.feature
+++ b/news/20201023.feature
@@ -1,0 +1,1 @@
+Add CMake variables for the supported application profiles to the generated CMake module
--- a/src/mbed_tools/build/_internal/cmake_file.py
+++ b/src/mbed_tools/build/_internal/cmake_file.py
@@ -60,6 +60,7 @@ def _render_mbed_config_cmake_template(
         "target_macros": target_build_attributes["macros"],
         "supported_form_factors": target_build_attributes["supported_form_factors"],
         "supported_c_libs": supported_c_libs,
+        "supported_application_profiles": target_build_attributes["supported_application_profiles"],
         "c_lib": target_build_attributes["c_lib"],
         "core": target_build_attributes["core"],
         "printf_lib": target_build_attributes["printf_lib"],
--- a/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
+++ b/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
@@ -15,6 +15,11 @@ list(APPEND MBED_TARGET_SUPPORTED_C_LIBS {% for supported_c_lib in supported_c_l
 {%- endfor %}
 )
 
+list(APPEND MBED_TARGET_SUPPORTED_APPLICATION_PROFILES {% for supported_application_profile in supported_application_profiles %}
+    {{supported_application_profile}}
+{%- endfor %}
+)
+
 list(APPEND MBED_TARGET_LABELS{% for label in labels %}
     {{label}}
 {%- endfor %}
--- a/tests/build/_internal/test_cmake_file.py
+++ b/tests/build/_internal/test_cmake_file.py
@@ -25,6 +25,7 @@ class TestGenerateCMakeListsFile(TestCase):
         mbed_target = "K64F"
         toolchain_name = "GCC"
         target["supported_c_libs"] = {toolchain_name.lower(): ["small", "std"]}
+        target["supported_application_profiles"] = ["full", "bare-metal"]
 
         result = generate_mbed_config_cmake_file(mbed_target, target, config, toolchain_name)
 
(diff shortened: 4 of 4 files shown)


hugueskamba:
@0xc0170 @rajkan01 
How does support for the `requires` feature work with the new tool?
What will happen if the user specifies other libraries as done [here](https://github.com/ARMmbed/mbed-os/blob/master/TESTS/configs/baremetal.json)? Will the additional libraries be added automatically by CMake? I am guessing "No". Is that a feature we also want to support? If not, do we need to document somewhere that adding libraries to `requires` has no effect with the new tools? Or do we want to document that `requires` is only used to select the baremetal profile?

0xc0170:
This is what we still need to discuss, lets talk on Monday and find a solution. My first attempt would be to display warning for requires with new tools (point to docs where the migration guide is - we could provide mapping there for the old components and new ones as I dont think they are 1:1, arent they?). `target_overrides` works as it did before, correct?

The new way is to use CMake target_link_libraries() instead of requires,  you can get the list of all components within Mbed OS (one thing we should check if it is already user friendly what CMake provides).

I would vote for migration path (should be made simple enough).

rajkan01:
> 
> 
> This is what we still need to discuss, lets talk on Monday and find a solution. My first attempt would be to display warning if `requires` is found in the configuration when using new tools (point to docs where the migration guide is - we could provide mapping there for the old components and new ones as I dont think they are 1:1, arent they?). `target_overrides` works as it did before, correct?
> 
> The new way is to use CMake target_link_libraries() instead of requires, you can get the list of all components within Mbed OS (one thing we should check if it is already user friendly what CMake provides).
> 
> I would vote for migration path (should be made simple enough): print a warning with docs url where the migration guide is, updating from json to CMake should be couple of lines.


I requested Mbedos-tools team and raised the Jira ticket to add the support for "requires" and override target attributes support (target.xxx) which is not supported. 
As you suggested new mbedos-tools to display a warning when `requires` is found in the configuration. once things get a better understanding we can slowly move to the new way is use CMake target_link_libraries()

src/mbed_tools/build/_internal/cmake_file.py
```diff
@@ -69,6 +71,7 @@ def _render_mbed_config_cmake_template(
         "macros": sorted(macros, key=lambda macro: macro.name),
         "max_name_length": max(_max_attribute_length(options, "macro_name"), _max_attribute_length(macros, "name")),
         "max_value_length": max(_max_attribute_length(options, "value"), _max_attribute_length(macros, "value")),
+        "baremetal_profile": "OFF"
```
hugueskamba, line 74: Could you explain how this can change to "ON"?

src/mbed_tools/build/_internal/cmake_file.py
```diff
@@ -50,6 +50,7 @@ def _render_mbed_config_cmake_template(
     options = list(config.options.values())
     macros = list(config.macros.values())
     supported_c_libs = [x for x in target_build_attributes["supported_c_libs"][toolchain_name.lower()]]
+    supported_application_profiles = [x for x in target_build_attributes["supported_application_profiles"]]
```
rwalton-arm, line 53: Why do we need to create this new list? Can't we just pass the `target_build_attributes["supported_application_profiles"]` straight to the `context`, as we have done for the other attributes?

rwalton-arm (changes requested):
Thanks for the PR.

The unit tests, linting and news file checks are all failing in CI. We need to get the CI passing before we can review this.

I also don't understand why need this PR if we're moving the bare metal/RTOS library selection to CMake using `target_link_libraries`. We should just implement it using `target_link_libraries` now instead of hard coding bare metal profile to "OFF" in the tools if it's something we're just going to remove at a later date.

src/mbed_tools/build/_internal/cmake_file.py
rajkan01: This flag ON/OFF based on "requires": "bare-metal" mbed_app.json config as like previous tools even I asked clarification in a mail sometimes ago.

mbedos-tools is yet to add support for "requires" config.

Patater (changes requested):
This needs a news file and unit tests.

hugueskamba:
> This needs a news file and unit tests.

Does the branch have to be in the upstream repository or can it be on a fork now?

src/mbed_tools/build/_internal/cmake_file.py
hugueskamba: Resolving this as `mbed-tools` will no longer provide a CMake variable to indicate whether the baremetal profile was selected or not.

Patater:
> > This needs a news file and unit tests.
> 
> Does the branch have to be in the upstream repository or can it be on a fork now?

Needs to be on upstream repo to have CI run still, unfortunately.

news/20201023.feature
```diff
@@ -0,0 +1 @@
+Add CMake variables for the supported application profiles to the generated CMake module
```
Patater, line 1: Add a new line to the end of this file, please.

Patater (changes requested):
This will need re-raising from a branch on the main repo instead of a fork, so we have CI coverage.

rajkan01:
I pushed changes dev/rajkan01/add_baremetal_profile_support branch aginst master branch in #93

PR #103, opened 2020-11-04
config: Add CMake variable for the path of the local Mbed OS program

hugueskamba:
`MBED_ROOT` is provided in the generated configuration
CMake module. It is used by application programs to add the subdirectory where the Mbed OS program is located.

### Test Coverage

### DIFF
--- a/src/mbed_tools/build/_internal/cmake_file.py
+++ b/src/mbed_tools/build/_internal/cmake_file.py
@@ -15,24 +15,27 @@ TEMPLATE_NAME = "mbed_config.tmpl"
 
 
 def generate_mbed_config_cmake_file(
-    mbed_target_name: str, target_build_attributes: dict, config: Config, toolchain_name: str
+    mbed_target_name: str, target_build_attributes: dict, config: Config, toolchain_name: str, mbed_os_path: pathlib.Path
 ) -> str:
     """Generate the top-level CMakeLists.txt file containing the correct definitions for a build.
 
     Args:
-        mbed_target: the target the application is being built for
-        target_build_attributes: map of target attributes
-        program_path: the path to the local Mbed program
-        toolchain_name: the toolchain to be used to build the application
+        mbed_target_name: the target the application is being built for.
+        target_build_attributes: map of target attributes.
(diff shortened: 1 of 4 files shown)


src/mbed_tools/build/_internal/cmake_file.py
```diff
@@ -15,24 +15,27 @@
 
 
 def generate_mbed_config_cmake_file(
-    mbed_target_name: str, target_build_attributes: dict, config: Config, toolchain_name: str
+    mbed_target_name: str, target_build_attributes: dict, config: Config, toolchain_name: str, mbed_os_path: str
```
rwalton-arm, line 18: `mbed_os_path` is a `pathlib.Path` and not a `str`. This is the reason `mypy` is failing in the CI.

rwalton-arm (changes requested):

Patater (changes requested):
"MBED_ROOT is provided in the generated configuration CMake module."

But why is this useful? Isn't it always relative to the application root at `mbed-os`? The commit message needs to explain the motivation for the change a bit more.

hugueskamba:
> "MBED_ROOT is provided in the generated configuration CMake module."
> 
> But why is this useful? Isn't it always relative to the application root at `mbed-os`? The commit message needs to explain the motivation for the change a bit more.

We supposed that users can decide to put `mbed-os` to a location other than the app root. One could put it in `<APP_NAME>/libs/mbed-os` . Would the tools generate the config file if ran at `<APP_NAME>`?

Currently every application [create a `MBED_ROOT` CMake variable](https://github.com/ARMmbed/mbed-os-example-blinky/blob/feature-cmake/CMakeLists.txt#L7), this PR creates it instead. However, if we think it is always guaranteed to be at the app root then applications can simple use `add_subdirectory(mbed-os)`.

@0xc0170 thoughts?

Patater:
> It is used by application programs to add the subdirectory where the Mbed OS program is located.

I still don't understand why application programs would need an MBED_ROOT variable to do this, as the application would know better than the tools would where the mbed-os folder is for add_subdirectory.

hugueskamba:
Won't change.

PR #110, opened 2020-11-09
Rename `checkout` to `deploy` command line option

jainvikas8:
### Description

To ensure good usability with the new Mbed tools the command line
options need to harmonize with the old tools.

Therefore change `checkout` to `deploy` by keeping the existing
functionality.

These are the following changes done:
* Docs updated
* Source files for project and CLI command updated to make sure api
calls mean the same externally and internally
* Unit tests updated with renamed test functions
* Travis job updated

### Test Coverage

### DIFF
--- a/src/mbed_tools/cli/project_management.py
+++ b/src/mbed_tools/cli/project_management.py
@@ -2,7 +2,7 @@
 # Copyright (C) 2020 Arm Mbed. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-"""Project management commands: new, clone, checkout and libs."""
+"""Project management commands: new, clone, deploy and libs."""
 import os
 import pathlib
 
@@ -11,7 +11,7 @@ from typing import Any
 import click
 import tabulate
 
-from mbed_tools.project import initialise_project, clone_project, get_known_libs, checkout_project_revision
+from mbed_tools.project import initialise_project, clone_project, get_known_libs, deploy_project
 
 
 @click.command()
@@ -74,17 +74,19 @@ def libs(path: str) -> None:
     click.echo(tabulate.tabulate(table, headers=headers))
 
     if lib_data["unresolved"]:
-        click.echo(
-            "\nUnresolved libraries detected. Please run the `checkout` command to download library source code."
-        )
+        click.echo("\nUnresolved libraries detected. Please run the `deploy` command to download library source code.")
 
 
 @click.command()
 @click.argument("path", type=click.Path(), default=os.getcwd())
 @click.option(
-    "--force", "-f", is_flag=True, show_default=True, help="Force checkout, overwrites local uncommitted changes."
+    "--force",
+    "-f",
+    is_flag=True,
+    show_default=True,
+    help="Forces checkout of all library repositories at specified commit in the .lib file, overwrites local changes.",
 )
-def checkout(path: str, force: bool) -> None:
+def deploy(path: str, force: bool) -> None:
     """Checks out Mbed program library dependencies at the revision specified in the ".lib" files.
 
     Ensures all dependencies are resolved and the versions are synchronised to the version specified in the library
@@ -95,4 +97,4 @@ def checkout(path: str, force: bool) -> None:
     REVISION: The revision of the Mbed project to check out.
     """
     click.echo("Checking out all libraries to revisions specified in .lib files. Resolving any unresolved libraries.")
-    checkout_project_revision(pathlib.Path(path), force)
+    deploy_project(pathlib.Path(path), force)
--- a/src/mbed_tools/project/_internal/libraries.py
+++ b/src/mbed_tools/project/_internal/libraries.py
@@ -70,7 +70,7 @@ class LibraryReferences:
         if list(self.iter_unresolved()):
             self.resolve()
 
-    def checkout(self, force: bool) -> None:
+    def deploy(self, force: bool) -> None:
         """Check out all resolved libs to revision specified in .lib files."""
         for lib in self.iter_resolved():
             repo = git_utils.init(lib.source_code_path)
--- a/src/mbed_tools/project/mbed_program.py
+++ b/src/mbed_tools/project/mbed_program.py
@@ -107,7 +107,7 @@ class MbedProgram:
         except ValueError as mbed_os_err:
             raise MbedOSNotFound(
                 f"Mbed OS was not found due to the following error: {mbed_os_err}"
-                "\nYou may need to resolve the mbed-os.lib reference. You can do this by performing a `checkout`."
+                "\nYou may need to resolve the mbed-os.lib reference. You can do this by performing a `deploy`."
             )
 
         return cls(program, mbed_os)
@@ -116,9 +116,9 @@ class MbedProgram:
         """Resolve all external dependencies defined in .lib files."""
         self.lib_references.resolve()
 
-    def checkout_libraries(self, force: bool = False) -> None:
+    def deploy_libraries(self, force: bool = False) -> None:
         """Check out all resolved libraries to revisions specified in .lib files."""
-        self.lib_references.checkout(force)
+        self.lib_references.deploy(force)
 
     def list_known_library_dependencies(self) -> List[MbedLibReference]:
         """Returns a list of all known library dependencies."""
--- a/src/mbed_tools/project/project.py
+++ b/src/mbed_tools/project/project.py
@@ -43,8 +43,8 @@ def initialise_project(path: pathlib.Path, create_only: bool) -> None:
         program.resolve_libraries()
 
 
-def checkout_project_revision(path: pathlib.Path, force: bool = False) -> None:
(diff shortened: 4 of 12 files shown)


src/mbed_tools/project/_internal/git_utils.py
```diff
@@ -41,18 +41,18 @@ def clone(url: str, dst_dir: Path) -> git.Repo:
...
-        repo: git.Repo object where the checkout will be performed.
+        repo: git.Repo object where the deploy will be performed.
         ref: Git commit hash, branch or tag reference, must be a valid ref defined in the repo.
 
     Raises:
         VersionControlError: Check out failed.
     """
     try:
-        repo.git.checkout(f"--force {ref}" if force else str(ref))
+        repo.git.deploy(f"--force {ref}" if force else str(ref))
```
rwalton-arm, line 55: This isn't going to work. `repo.git.checkout` is a call to the `gitpython` library. The parent function doesn't make sense as `deploy` either, this is a `git_utils` module and `git` doesn't have a `deploy` command. We should only need to change the CLI command handler and external name as that's the only place `deploy` makes sense.

src/mbed_tools/project/_internal/libraries.py
```diff
@@ -64,19 +64,19 @@ def resolve(self) -> None:
...
                 logger.info(f"Checking out revision {git_ref.ref} for library {git_ref.repo_url}.")
-                git_utils.checkout(repo, git_ref.ref)
+                git_utils.deploy(repo, git_ref.ref)
 
         # Check if we find any new references after cloning dependencies.
         if list(self.iter_unresolved()):
             self.resolve()
 
-    def checkout(self, force: bool) -> None:
+    def deploy(self, force: bool) -> None:
```
rwalton-arm, line 73: This is an internal API, I wouldn't bother changing the name here.

src/mbed_tools/project/mbed_program.py
```diff
@@ -116,9 +116,9 @@ def resolve_libraries(self) -> None:
         """Resolve all external dependencies defined in .lib files."""
         self.lib_references.resolve()
 
-    def checkout_libraries(self, force: bool = False) -> None:
+    def deploy_libraries(self, force: bool = False) -> None:
```
rwalton-arm, line 119: `deploy_libraries` doesn't make sense. We should just leave the internal APIs and only change the command name and handler.

src/mbed_tools/project/project.py
```diff
@@ -43,20 +43,20 @@ def initialise_project(path: pathlib.Path, create_only: bool) -> None:
         program.resolve_libraries()
 
 
-def checkout_project_revision(path: pathlib.Path, force: bool = False) -> None:
-    """Checkout a specific revision of the current Mbed project.
+def deploy_project_revision(path: pathlib.Path, force: bool = False) -> None:
```
rwalton-arm, line 46: `deploy_project_revision` doesn't make sense. We should just leave the API names how they are, or at least come up with something that makes sense.

src/mbed_tools/project/_internal/libraries.py
jainvikas8: Idea is to keep internal and external in sync to avoid confusion

rwalton-arm: I don't think it is confusing. Internal APIs should be describing "how we are doing it", and we are doing the deploy by using git checkout. We should just change the command names and the click callback names as they are really the only place the "deploy" terminology makes sense.

README.md
```diff
@@ -16,7 +16,7 @@
 ## Overview
 
 This is the **future** command line tool for Mbed OS. It provides the ability to detect Mbed Enabled devices connected
-by USB, checkout Mbed projects and perform builds amongst other operations.
+by USB, deploy Mbed projects and perform builds amongst other operations.
```
rwalton-arm, line 19: We need to change this wording to something that makes sense. As a reader of this I'm not sure where or how we're deploying the mbed project, or what deploying even means in the context of an mbed project.

src/mbed_tools/project/project.py
rwalton-arm: Given that this is the public API we could change the name to `deploy` here. We should definitely change it from `deploy_project_revision` as that doesn't make sense to a reader of this code.

rwalton-arm (changes requested):

src/mbed_tools/project/_internal/libraries.py
jainvikas8: I'm not clear on your thoughts here, please elaborate with an example.

src/mbed_tools/project/project.py
jainvikas8: I think it's the same as 'deploy a project revision' or 'checkout a project revision'. Naming it to `deploy` would create confusion as there is already a definition/function with the same.

rwalton-arm: With your changes we already have multiple functions and methods named deploy. They are all namespaced in their own modules though.

The function names before and after your changes:
```
cli.checkout -> cli.deploy
project.checkout_project_revision -> project.deploy_project_revision
project.MbedProgram.checkout_libraries -> project.MbedProgram.deploy_libraries
project._internal.LibraryReferences.checkout -> project._internal.LibraryReferences.deploy
```
So you can see you already have multiple functions named deploy now (project._internal.LibraryReferences.deploy & cli.deploy). So, I'm not sure I'm getting your point about "creating confusion". All I know is the name `deploy_project_revision` is already creating confusion for me as it doesn't make much sense as a phrase.

src/mbed_tools/cli/project_management.py
```diff
@@ -74,17 +74,15 @@ def libs(path: str) -> None:
...
-            "\nUnresolved libraries detected. Please run the `checkout` command to download library source code."
-        )
+        click.echo("\nUnresolved libraries detected. Please run the `deploy` command to download library source code.")
 
 
 @click.command()
 @click.argument("path", type=click.Path(), default=os.getcwd())
 @click.option(
-    "--force", "-f", is_flag=True, show_default=True, help="Force checkout, overwrites local uncommitted changes."
+    "--force", "-f", is_flag=True, show_default=True, help="Force deploy, overwrites local uncommitted changes."
```
rwalton-arm, line 83: I think the help should say something like "Forces all library repositories to be checked out at the commit specified in the .lib file, overwrites local changes."

tests/project/_internal/test_libraries.py
```diff
@@ -49,57 +49,57 @@ def test_hydrates_recursive_dependencies(self, mock_clone, fs):
     @patchfs
     @mock.patch("mbed_tools.project._internal.git_utils.checkout", autospec=True)
     @mock.patch("mbed_tools.project._internal.git_utils.init", autospec=True)
-    def test_does_not_perform_checkout_if_no_git_ref_exists(self, mock_init, mock_checkout, mock_clone, fs):
+    def test_does_not_perform_deploy_if_no_git_ref_exists(self, mock_init, mock_deploy, mock_clone, fs):
```
rwalton-arm, line 52: I mentioned this in our call but forgot to comment here; all the tests in this file mock out git operations from `git_utils`, I don't think we need to rename them to deploy.

jainvikas8:
Forced pushed to address @rwalton-arm  comments.

jainvikas8:
Rebased to resolve conflicts

src/mbed_tools/cli/project_management.py
```diff
@@ -74,17 +74,22 @@ def libs(path: str) -> None:
...
 @click.command()
 @click.argument("path", type=click.Path(), default=os.getcwd())
 @click.option(
-    "--force", "-f", is_flag=True, show_default=True, help="Force checkout, overwrites local uncommitted changes."
+    "--force",
+    "-f",
+    is_flag=True,
+    show_default=True,
+    help="""
+            Forces all library repositories to be checked out at the commit specified in the .lib file,
```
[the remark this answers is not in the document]
jainvikas8: The reason this was written this way because of formatting limit of 120 characters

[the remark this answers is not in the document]
jainvikas8: Reworded it to make sure it fits... if not will rework again.

jainvikas8:
Forced pushed to address comments from @Patater

40Grit:
Will the tools maintain a hash of the sources for the project and all imported dependencies?

These commands were near worthless in Mbed-cli because they gave no guarantee that the sources in the project has not been modified. Nor were there commands to check if dependency sources had been changed.

jainvikas8:
> Will the tools maintain a hash of the sources for the project and all imported dependencies?
> 
> These commands were near worthless in Mbed-cli because they gave no guarantee that the sources in the project has not been modified. Nor were there commands to check if dependency sources had been changed.

I'm unsure with your comment here, please could you elaborate with an example or use case scenario?

rwalton-arm (approved):

PR #114, opened 2020-11-11
Serial Terminal

rwalton-arm:
### Description
Implement serial terminal and sterm command.

### Test Coverage

### DIFF
--- a/tests/sterm/test_terminal.py
+++ b/tests/sterm/test_terminal.py
@@ -1,0 +1,225 @@
+#
+# Copyright (C) 2020 Arm Mbed. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+from unittest import mock
+
+import sys
+
+import pytest
+
+from mbed_tools.sterm import terminal
+
+
+@pytest.fixture
+def mock_serial():
+    with mock.patch("mbed_tools.sterm.terminal.Serial") as serial:
+        yield serial
+
+
+@pytest.fixture
+def mock_console():
+    with mock.patch("serial.tools.miniterm.Console") as console:
+        yield console
+
+
+@pytest.fixture
+def mock_sterm():
+    with mock.patch("mbed_tools.sterm.terminal.SerialTerminal") as sterm:
+        yield sterm
+
+
+def test_initialises_serial_port(mock_sterm, mock_serial):
+    port = "tty.1111"
+    baud = 9600
+
+    terminal.run(port, baud)
+
+    mock_serial.assert_called_once_with(port=port, baudrate=str(baud))
+
+
+def test_initialises_sterm(mock_sterm, mock_serial):
+    port = "tty.1111"
+    baud = "9600"
+
+    terminal.run(port, baud)
+
+    mock_sterm.assert_called_once_with(mock_serial(), echo=True)
+
+
+def test_starts_sterm_thread(mock_sterm, mock_serial):
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().start.assert_called_once()
+
+
+def test_joins_tx_and_rx_threads(mock_sterm, mock_serial):
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().join.assert_any_call(True)
+
+
+def test_joins_tx_thread_after_keyboard_interrupt(mock_sterm, mock_serial):
+    mock_sterm().join.side_effect = (KeyboardInterrupt(), None)
+
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().join.assert_called_with()
+
+
+def test_closes_sterm(mock_sterm, mock_serial):
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().close.assert_called_once()
+
+
+def test_closes_sterm_after_exception(mock_sterm, mock_serial):
+    mock_sterm().join.side_effect = (Exception(), None)
+    with pytest.raises(Exception):
+        terminal.run("tty.122", 9600)
+
+    mock_sterm().close.assert_called_once()
+
+
+def test_closes_sterm_after_keyboard_interrupt(mock_sterm, mock_serial):
+    mock_sterm().join.side_effect = (KeyboardInterrupt(), None)
+    terminal.run("tty.122", 9600)
+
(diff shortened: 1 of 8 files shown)


tests/sterm/test_terminal.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+@pytest.mark.parametrize("menu_key", INVALID_MENU_KEYS)
+def test_ignores_invalid_menu_key(menu_key, mock_serial, mock_console):
+    term = terminal.SerialTerminal(mock_serial())
+    term.handle_menu_key = mock.Mock()
+    term.alive = True
+    mock_console().getkey.side_effect = (terminal.CTRL_T, menu_key)
+
+    with pytest.raises(StopIteration):
```
Patater, line 142: Why is a StopIteration an invalid menu key? Why not KeyboardError or something?

tests/sterm/test_terminal.py
```diff
@@ -0,0 +1,213 @@
...
+
+    term.handle_menu_key.assert_not_called()
+
+
+def test_reset_sends_serial_break(mock_serial, mock_console):
+    term = terminal.SerialTerminal(mock_serial())
+
+    term.reset()
+
+    mock_serial().sendBreak.assert_called_once()
```
Patater, line 153: Do we really need to send break conditions? Is this a reliable way to reset a target? @jamesbeyond Do you know if we have any targets that reset when given a serial break? Is this a common thing?

tests/sterm/test_terminal.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def test_ctrl_b_sends_reset_to_serial_port(mock_serial, mock_console):
+    term = terminal.SerialTerminal(mock_serial())
+    term.alive = True
+    mock_console().getkey.side_effect = (terminal.CTRL_B, terminal.CTRL_C)
```
Patater, line 159: Would this test pass if we did nothing for CTRL_B and made CTRL_C send a break to the serial port? How do we know it's the CTRL_B that's giving us the desired effect and not something else? I am curious why we sent CTRL_C. I guess it is to cause the console to stop blocking, but perhaps there is a more targeted way to do that.

tests/sterm/test_terminal.py
```diff
@@ -0,0 +1,213 @@
...
+def test_starts_sterm_thread(mock_sterm, mock_serial):
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().start.assert_called_once()
+
+
+def test_joins_tx_and_rx_threads(mock_sterm, mock_serial):
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().join.assert_any_call(True)
```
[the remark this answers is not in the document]
rwalton-arm: `join(True)` means "join the tx and rx threads" - `Miniterm` handles the thread management after that so it seemed good enough to just check we called `join` correctly here.

tests/sterm/test_terminal.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def test_joins_tx_thread_after_keyboard_interrupt(mock_sterm, mock_serial):
+    mock_sterm().join.side_effect = (KeyboardInterrupt(), None)
+
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().join.assert_called_with()
```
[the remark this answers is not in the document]
rwalton-arm: No they aren't equivalent,
 `assert_any_call` means "if the mock was ever called with `args`"
 `assert_called_with` means "if the most recent call was with `args`"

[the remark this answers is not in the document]
rwalton-arm: We check sterm closes after any exception below.

tests/sterm/test_terminal.py
```diff
@@ -0,0 +1,213 @@
...
+def test_joins_tx_and_rx_threads(mock_sterm, mock_serial):
+    terminal.run("tty.122", 9600)
+
+    mock_sterm().join.assert_any_call(True)
+
+
+def test_joins_tx_thread_after_keyboard_interrupt(mock_sterm, mock_serial):
+    mock_sterm().join.side_effect = (KeyboardInterrupt(), None)
+
+    terminal.run("tty.122", 9600)
```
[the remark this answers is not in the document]
rwalton-arm: `KeyboardInterrupt` is handled by `terminal.run` so it doesn't bubble up and fail the test.

tests/sterm/test_terminal.py
rwalton-arm: Not sure, this was just what the old tool did. Serial break does reset my DISCO_L475 though.

tests/sterm/test_terminal.py
rwalton-arm: CTRL_C doesn't send a break to the serial port, it tells the terminal to stop running and blocking. Maybe it's better to just handle the `StopIteration` like I did in the above test, to make it clearer we're not testing CTRL_C handling. I did think the `StopIteration` handling was a bit dirty though...

tests/sterm/test_terminal.py
rwalton-arm: `StopIteration` isn't a valid menu key, it's a way to get the terminal to stop blocking so I can assert on the state of the mock.

tests/sterm/test_terminal.py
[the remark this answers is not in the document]
rwalton-arm: Actually, below we only check `Exception` derived exceptions. I'll add a test for close on `KeyboardInterrupt`.

tests/sterm/test_terminal.py
rwalton-arm: Note: If we didn't send serial break and change the menu text and key handling, we wouldn't need to subclass `Miniterm` at all. However, we will change the UX if we do that.

Patater: Alright. Learn a new feature everyday. If we run --sterm without --flash, I can see it being useful that the target would be reset. I would have expected a separate reset mechanism than a serial break condition, but it works how it works.

Patater (approved):

jainvikas8 (approved):

PR #117, opened 2020-11-16
Fix deploy command for examples containing multiple apps 

rwalton-arm:
### Description
The deploy command should be able to fetch libraries in projects containing multiple mbed application subprojects.

This PR also removes the `libs` command and instead prints a table of fetched libraries after the user runs `import` or `deploy`.

### Test Coverage


tests/project/test_mbed_project.py
```diff
@@ -26,9 +39,8 @@ def test_skips_mbed_os_when_create_only_is_true(self, mock_program):
         mock_program.from_new.return_value.resolve_libraries.assert_not_called()
 
 
-@mock.patch("mbed_tools.project.project.MbedProgram", autospec=True)
-class TestCloneProject(TestCase):
-    def test_clones_from_remote(self, mock_program):
+class TestCloneProject:
+    def test_clones_from_remote(self, mock_git):
```
Patater, line 49: `test_clones_from_remote()` used to use `mock_program`. In commit "mbed_project: Remove unittest.TestCase from tests", do we want to move it to use `mock_git` as is done here, or should we keep it using `mock_program`? This probably should use `mock_program` if we want this commit to pass tests.

tests/project/_internal/test_git_utils.py
```diff
@@ -24,17 +24,34 @@ def mock_progress():
...
+
+        with pytest.raises(VersionControlError, match="exists and is not an empty directory"):
+            git_utils.clone("https://blah", dst_dir)
+
+    def test_can_clone_to_empty_existing_dst_dir(self, mock_repo, tmp_path, mock_progress):
+        dst_dir = Path(tmp_path, "dst")
+        dst_dir.mkdir()
+        url = "https://repo"
+
+        git_utils.clone(url, dst_dir)
```
Patater, line 52: Should we `assert repo is not None` or is this completely covered fine by `test_returns_repo()`? I see the test here is slightly different, in that we've created the dst_dir before calling clone. We might assume that the return from clone will not be None given no exceptions happened, but shouldn't we assert our assumptions? I understand this might be a bit of a stylistic choice. Let me know what you think best here.

src/mbed_tools/project/project.py
```diff
@@ -55,11 +59,11 @@ def deploy_project(path: pathlib.Path, force: bool = False) -> None:
         force: Force overwrite uncommitted changes. If False, the deploy will fail if there are uncommitted local
```
Patater, line 61: Pre-existing issue: There is no `project_revision` parameter, so it's a bit odd to attempt documenting it here. Should fix in separate commit.

tests/project/test_mbed_project.py
```diff
@@ -24,55 +30,56 @@ def mock_git():
...
+    def test_fetches_mbed_os_when_create_only_is_false(self, mock_libs, mock_program):
         path = pathlib.Path()
         initialise_project(path, create_only=False)
 
         mock_program.from_new.assert_called_once_with(path)
-        mock_program.from_new.return_value.resolve_libraries.assert_called_once()
+        mock_libs().fetch.assert_called_once()
 
-    def test_skips_mbed_os_when_create_only_is_true(self, mock_program):
+    def test_skips_mbed_os_when_create_only_is_true(self, mock_program, mock_libs):
```
Patater, line 40: nitpick: Could we keep the order of mock_program and mock_libs consistent between tests (all in this file with similar shape)?

src/mbed_tools/project/mbed_program.py
```diff
@@ -6,18 +6,16 @@
 import logging
 
 from pathlib import Path
-from typing import List, Dict
+from typing import Dict
```
Patater, line 9: In commit "MbedProgram: Remove library handling", please explain the motivation behind the removal (based on the changes contributed in previous commits).

src/mbed_tools/cli/main.py
```diff
@@ -15,7 +15,7 @@
 
 from mbed_tools.cli.configure import configure
 from mbed_tools.cli.list_connected_devices import list_connected_devices
-from mbed_tools.cli.project_management import new, clone, deploy, libs
+from mbed_tools.cli.project_management import new, clone, deploy
```
Patater, line 18: Commit "project: Remove libs command" creates a need for another PR to mbed-os-5-docs to update the documentation on "how to list dependencies": https://github.com/ARMmbed/mbed-os-5-docs/blob/development/docs/tools/mbed_cli_2/use.md#list-an-applications-library-dependencies

src/mbed_tools/cli/project_management.py
```diff
@@ -97,4 +81,24 @@ def deploy(path: str, force: bool) -> None:
...
+        table.append(
+            [
+                lib.reference_file.stem,
+                lib.get_git_reference().repo_url,
+                lib.source_code_path,
+                lib.get_git_reference().ref,
+            ]
+        )
+
+    headers = ("Library Name", "Repository URL", "Location", "Git Reference")
```
Patater, line 103: Do you think "Path" is a more clear name than "Location"? To me, "Location" could mean remote location, if it weren't for the URL also being listed here.

src/mbed_tools/project/project.py
```diff
@@ -79,4 +81,4 @@ def get_known_libs(path: pathlib.Path) -> Dict[str, Any]:
         detected.
```
Patater, line 81: The `Returns` documentation needs updating, as it says a dictionary is returned when we now return a list.

src/mbed_tools/project/project.py
```diff
@@ -33,6 +33,8 @@ def clone_project(url: str, dst_path: Any = None, recursive: bool = False) -> No
         libs = LibraryReferences(root=dst_path, ignore_paths=["mbed-os"])
```
Patater, line 33: We now return something, so the docstring should have some `Returns:` documentation.

Patater (changes requested):

src/mbed_tools/project/project.py
rwalton-arm: This was fixed in ff5b5a265a972e2906655dbe6ef7ee9c9902f4ab. I need to rebase.

src/mbed_tools/cli/main.py
rwalton-arm: Added a Jira task to update the docs.

tests/project/_internal/test_git_utils.py
rwalton-arm: It is technically covered by `test_returns_repo`. However, we should really assert on the return value here, even if it's just for consistency with `test_returns_repo`, as these two test cases are verifying the "success case", so we should verify the postconditions in both cases.

Patater:
Dropping the `libs` command should have a .major news file.

tests/cli/test_project_management.py
```diff
@@ -4,44 +4,65 @@
...
+
+@pytest.fixture
+def mock_deploy_project():
+    with mock.patch("mbed_tools.cli.project_management.deploy_project") as deploy_proj:
+        yield deploy_proj
+
+
+@pytest.fixture
+def mock_get_libs():
+    with mock.patch("mbed_tools.cli.project_management.get_known_libs") as deploy_proj:
```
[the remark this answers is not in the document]
rwalton-arm: Fixed.

tests/cli/test_project_management.py
```diff
@@ -4,44 +4,65 @@
...
         CliRunner().invoke(clone, ["url", "dst"])
-        mocked_clone_project.assert_called_once_with("url", pathlib.Path("dst"), True)
+        mock_clone_project.assert_called_once_with("url", pathlib.Path("dst"), True)
 
 
-@mock.patch("mbed_tools.cli.project_management.get_known_libs", autospec=True)
 class TestLibsCommand(TestCase):
     def test_calls_libs_function(self, mocked_get_libs):
         CliRunner().invoke(libs)
         mocked_get_libs.assert_called_once()
```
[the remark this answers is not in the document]
rwalton-arm: Fixed.

tests/project/test_mbed_project.py
rwalton-arm: Fixed.

tests/project/_internal/test_git_utils.py
rwalton-arm: Asserted.

src/mbed_tools/project/project.py
rwalton-arm: Rebased.

tests/project/test_mbed_project.py
rwalton-arm: Fixed.

src/mbed_tools/cli/project_management.py
rwalton-arm: Chenged name to "Path".

src/mbed_tools/project/project.py
rwalton-arm: Documented.

src/mbed_tools/project/project.py
rwalton-arm: Documented.

rwalton-arm:
> Dropping the `libs` command should have a .major news file.

Added.

src/mbed_tools/project/mbed_program.py
rwalton-arm: Added an explanation to the commit message.

tests/project/test_mbed_project.py
```diff
@@ -26,8 +39,7 @@ def test_skips_mbed_os_when_create_only_is_true(self, mock_program):
         mock_program.from_new.return_value.resolve_libraries.assert_not_called()
 
 
-@mock.patch("mbed_tools.project.project.MbedProgram", autospec=True)
-class TestImportProject(TestCase):
```
jainvikas8, line 30: Do we need to rename this class?

tests/project/test_mbed_project.py
```diff
@@ -42,8 +54,7 @@ def test_resolves_libs_when_recursive_is_true(self, mock_program):
         mock_program.from_url.return_value.resolve_libraries.assert_called_once()
 
 
-@mock.patch("mbed_tools.project.project.MbedProgram", autospec=True)
-class TestDeployProject(TestCase):
```
jainvikas8, line 46: Do we need to rename this class?

tests/cli/test_project_management.py
```diff
@@ -4,44 +4,65 @@
...
 
-        self.assertEqual(
-            result.output,
-            "Creating a new Mbed program at path 'path'.\nDownloading mbed-os and adding it to the project.\n",
-        )
+        assert result.output == expected
 
 
-@mock.patch("mbed_tools.cli.project_management.import_project", autospec=True)
-class TestImportCommand(TestCase):
```
jainvikas8, line 30: Do we need to rename the class here?

tests/project/test_mbed_project.py
rwalton-arm: Rebase fail. I'll fix it up.

jainvikas8 (approved):

src/mbed_tools/project/project.py
Patater: I still see `project_revision: Revision of the Mbed project to check out.` in `project.py`.

rwalton-arm: Oh, sorry it wasn't fixed upstream. I've fixed it in 1dd5e7b989339817cacaaf1ab4a0757cec35cb23

Patater (approved):
LGTM

Issue #122, opened 2020-11-23
deploy: cannot update to latest version of a library by modifying `.lib`


urutva:
### Description

`mbedtools deploy` command fails to fetch latest version of library after `.lib` is updated.

Steps to reproduce:
* mbedtools clone mbed-os-example-blinky
* cd mbed-os-example-blinky
* Content of `mbed-os.lib` (mbed-os-6.4.0) --> https://github.com/ARMmbed/mbed-os/#8ef0a435b2356f8159dea8e427b2935d177309f8
* mbedtools deploy
* update mbed-os.lib to point to latest master --> https://github.com/ARMmbed/mbed-os/
* mbedtools deploy

The version of mbed-os is still 8ef0a435b2356f8159dea8e427b2935d177309f8 (mbed-os-6.4.0) and not f2278567d09b9ae9f4843e1d9d393526b9462783 (mbed-os-6.5.0).

### Issue request type

- [ ] Enhancement
- [x] Bug

PR #128, opened 2020-11-26
Fetch and checkout default branch if no git hash given in .lib file

jainvikas8:
### Description

Set default branch to master
Make sure we checkout `master` if there is no git hash present in the .lib file.

Perform fetch before checkout
Make sure we fetch git repo before performing checkout when using `deploy` command. This makes sure we have the latest changes in repo.

Tests have been updated to accommodate this issue.

### Test Coverage

Fixes https://github.com/ARMmbed/mbed-tools/issues/122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -157,3 +157,17 @@ matrix:
         - mbedtools deploy
         - mbedtools compile -t GCC_ARM -m K64F
         - ccache -s
+
+    - <<: *mbed-tools-test
+      name: "Test deploy command checks out tip of default branch"
+      env: NAME=test-deploy-cmd CACHE_NAME=test-deploy-cmd
+      script:
+        - mkdir deploy-test
+        - cd deploy-test
+        - echo "https://github.com/ARMmbed/mbed-os" > mbed-os.lib
+        - git clone --branch=latest https://github.com/ARMmbed/mbed-os.git
+        - mbedtools -vvv deploy
+        - cd mbed-os
+        - git branch --points-at origin/HEAD
+        - git branch --points-at origin/HEAD | grep -q "\*" || (echo "Couldn't
+          see current branch points at default branch" && false)
--- a/src/mbed_tools/cli/project_management.py
+++ b/src/mbed_tools/cli/project_management.py
@@ -12,6 +12,7 @@ import click
 import tabulate
 
 from mbed_tools.project import initialise_project, import_project, get_known_libs, deploy_project
+from mbed_tools.project._internal import git_utils
 
 
 @click.command()
@@ -93,7 +94,9 @@ def _print_dependency_table(libs: List) -> None:
                 lib.reference_file.stem,
                 lib.get_git_reference().repo_url,
                 lib.source_code_path,
-                lib.get_git_reference().ref,
+                git_utils.get_default_branch(git_utils.get_repo(lib.source_code_path))
+                if not lib.get_git_reference().ref
+                else lib.get_git_reference().ref,
             ]
         )
 
--- a/src/mbed_tools/project/_internal/git_utils.py
+++ b/src/mbed_tools/project/_internal/git_utils.py
@@ -96,3 +96,21 @@ def get_repo(path: Path) -> git.Repo:
         raise VersionControlError(
             "Could not find a valid git repository at this path. Please perform a `git init` command."
         )
+
+
+def get_default_branch(repo: git.Repo) -> str:
+    """Get a default branch from an existing git.Repo.
+
+    Args:
+        repo: git.Repo object
+
+    Returns:
+        The default branch name as a string.
+
+    Raises:
+        VersionControlError: Could not find the default branch name.
+    """
+    try:
+        return str(repo.git.symbolic_ref("refs/remotes/origin/HEAD").rsplit("/", maxsplit=1)[-1])
+    except git.exc.GitCommandError as err:
+        raise VersionControlError(f"Could not resolve default repository branch name. Error from VCS: {err}")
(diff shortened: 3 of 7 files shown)


src/mbed_tools/cli/project_management.py
```diff
@@ -93,7 +93,7 @@ def _print_dependency_table(libs: List) -> None:
                 lib.reference_file.stem,
                 lib.get_git_reference().repo_url,
                 lib.source_code_path,
-                lib.get_git_reference().ref,
+                "master" if lib.get_git_reference().ref == "" else lib.get_git_reference().ref,
```
Patater, line 96: I don't believe we need to nor should do this. We should clone without specifying a reference when the .lib file has no reference, in order to use the actual default branch for the repo to clone: this may or may not be master. Trying to clone master will fail on repositories that do not use `master` for any branch, and using `master` will not clone the default branch when the default is something other than `master` (as determined by the remote repository).

src/mbed_tools/project/_internal/libraries.py
```diff
@@ -62,9 +62,12 @@ def fetch(self) -> None:
             git_ref = lib.get_git_reference()
             logger.info(f"Resolving library reference {git_ref.repo_url}.")
             repo = git_utils.clone(git_ref.repo_url, lib.source_code_path)
-            if git_ref.ref:
-                logger.info(f"Checking out revision {git_ref.ref} for library {git_ref.repo_url}.")
-                git_utils.checkout(repo, git_ref.ref)
+
+            if git_ref.ref == "":
+                git_ref.ref = "master"
```
Patater, line 67: Don't hard-code master here. It can easily be wrong. If no ref is given for `fetch()`, clone without doing a checkout. If you are trying to make the local checkout match the latest default branch upstream, you need a different approach.

Patater (changes requested):

src/mbed_tools/project/_internal/libraries.py
jainvikas8: How can we query the "default" branch in a repo?

Patater: When cloning, no need to query the default branch. The clone will grab the default branch naturally.

When doing a checkout, you may be able to use `git checkout origin/HEAD` to move to the correct git hash, but this does not create a local branch tracking the remote default branch.

.travis.yml
```diff
@@ -157,3 +157,14 @@ matrix:
         - mbedtools deploy
         - mbedtools compile -t GCC_ARM -m K64F
         - ccache -s
+
+    - <<: *mbed-tools-test
+      name: "Test deploy command"
```
Patater, line 162: Could you explain what this test is doing for the "name"?

It looks like you are not testing deploy how the user would use it (as you still clone mbed-os manually with git).

Patater (changes requested):
In commit message for "fix: Use default branch from remote repository", please add a line by itself that says "Fixes #122".

rwalton-arm (changes requested):

src/mbed_tools/project/_internal/libraries.py
```diff
@@ -75,8 +75,11 @@ def checkout(self, force: bool) -> None:
         for lib in self.iter_resolved():
             repo = git_utils.get_repo(lib.source_code_path)
             git_ref = lib.get_git_reference()
-            if git_ref.ref:
-                git_utils.checkout(repo, git_ref.ref, force=force)
+            if git_ref.ref == "":
+                git_ref.ref = git_utils.get_default_branch(repo)
+
+            repo.git.fetch()
```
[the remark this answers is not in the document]
jainvikas8: yes

.travis.yml
jainvikas8: Idea is to clone some version of Mbed OS to unknown commit and ask `mbed deploy` to update the libraries

.travis.yml
```diff
@@ -157,3 +157,14 @@ matrix:
...
+      env: NAME=test-deploy-cmd CACHE_NAME=test-deploy-cmd
+      script:
+        - git clone --depth=1 --single-branch --branch development https://github.com/ARMmbed/mbed-os-example-blinky.git
+        - cd mbed-os-example-blinky
+        - git clone --depth=1 --branch mbed-os-6.1.0 https://github.com/ARMmbed/mbed-os.git
```
rwalton-arm, line 167: This test won't work, as we clone only a single branch at depth=1 and then try to fetch other branches. Git doesn't allow that. We'll have to come up with a better test case here.

jainvikas8: Removed the depth, as we still looking to `mbed deploy` with an empty hash

src/mbed_tools/project/_internal/git_utils.py
```diff
@@ -96,3 +96,21 @@ def get_repo(path: Path) -> git.Repo:
...
+    Returns:
+        The default branch name as a string.
+
+    Raises:
+        VersionControlError: Could not find the default branch name.
+    """
+    try:
+        return str(repo.git.symbolic_ref("refs/remotes/origin/HEAD").rsplit("/", maxsplit=1)[-1])
+    except git.exc.GitCommandError:
+        raise VersionControlError("Could not resolve default repository branch name")
```
rwalton-arm, line 116:
```suggestion
    except git.exc.GitCommandError as err:
        raise VersionControlError(f"Could not resolve default repository branch name. Error from VCS: {err}")
```

This means we will propagate the error message from `git`, which may be useful information for the user.

.travis.yml
Patater: Explain it in the name field please, not on the PR.

Patater: I've described it.

rwalton-arm (approved):

PR #130, opened 2020-11-27
ci: Remove _clone_dependencies from travis CI

### LINKED ISSUE #126: Remove `_clone_dependencies` from CI
### Description

[`_clone_dependencies`](https://github.com/ARMmbed/mbed-tools/blob/master/travis-ci/functions.sh#L73) was introduced to support mbed-os examples that have sub-example folder. But the latest `mbed-tools import` ( >3.5.0) command now supports examples with sub-example folder.  `_clone_dependencies` functionality should be replaced by `mbed-tools import` command.

### Issue request type

- [x] Enhancement
- [ ] Bug

rwalton-arm:
### Description
Previously mbed-tools checked for the presence of an mbed-os.lib file
before performing an `import` or `deploy`. We worked around this in
our CI by creating an empty mbed-os.lib at the root of the project
before running the `deploy` command. mbed-tools no longer enforces this
policy, so we can remove the workaround.

After removing the workaround there seems to be no need for the
`_clone_dependencies` function. This commit removes the function. We
can't replace it with `mbedtools import` as that would cause many
unneeded copies of mbed-os to be cloned in the case of the BLE example,
which would slow the CI down significantly. For this reason we keep the
git clone step for the example and run `mbedtools deploy` in the
example/subexample folder to fetch other dependencies.

Fixes #126.

### Test Coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,9 +66,17 @@ matrix:
         # Leave mbed-tools directory to clone dependencies
         - cd ..
       script:
-        - _clone_dependencies
+          # We use manual clone, with depth and single branch = the fastest
+        - git clone --depth=1 --single-branch --branch development https://github.com/ARMmbed/${EXAMPLE_NAME}.git
+        - |
+          if [ -z ${SUBEXAMPLE_NAME} ]; then
+              cd ${EXAMPLE_NAME}
+          else
+              cd ${EXAMPLE_NAME}/${SUBEXAMPLE_NAME}
+          fi
+        - mbedtools -vvv deploy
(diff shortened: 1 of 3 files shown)


rwalton-arm:
CI is failing on the "Test Cellular Example (WIO_3G)" - looking into this failure now.

jainvikas8:
Should we just resolve this https://github.com/ARMmbed/mbed-tools/issues/126 by only using - `mbed import`  instead? Merging https://github.com/ARMmbed/mbed-tools/pull/131 should aid this too.

Patater (approved):

urutva (approved):

PR #135, opened 2020-11-30
ci: Add nightly release pipeline

rwalton-arm:
### Description
Add a pipeline to automate checking in auto generated files and tagging a new release. 

### Test Coverage

### DIFF
--- a/azure-pipelines/build-and-deploy.yml
+++ b/azure-pipelines/build-and-deploy.yml
@@ -33,10 +33,6 @@ stages:
                 - script: |
                     python -m pip install --upgrade tox
                     tox -e py37
-                  displayName: "Run pytests"
-
-                - script: |
-                    python -m pip install --upgrade tox
                     tox -e pypirelease -- -p $TWINE_PASSWORD -u $TWINE_USERNAME
                   displayName: "PyPI release"
                   env:
--- a/azure-pipelines/build-and-deploy.yml
+++ b/azure-pipelines/build-and-deploy.yml
@@ -11,6 +11,7 @@ pr:
 
 stages:
   - stage: ReleaseToPyPI
+    condition: not(and(eq(variables['Build.SourceBranch'], 'refs/heads/master'), contains(variables['Build.SourceVersionMessage'], 'Prepare mbed-tools release')))
     displayName: 'Production Release'
     jobs:
       - deployment: PyPIRelease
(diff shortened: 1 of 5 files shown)


jainvikas8 (approved):

azure-pipelines/prep-release.yml
```diff
@@ -0,0 +1,39 @@
...
+          git config --global user.name "Monty Bot"
+          git config --global user.email "monty-bot@arm.com"
+
+          python -m pip install --upgrade tox
+
+          git checkout master
+          tox -e preprelease
+          git show
+          git push https://%GIT_TOKEN%@github.com/ARMmbed/mbed-tools.git master
+          git push https://%GIT_TOKEN%@github.com/ARMmbed/mbed-tools.git --tags
```
Patater, line 36: We don't want to push the tag until the pre-release job is finished running, otherwise the pre-release job will see the tag after it finishes testing and will push a real release version.

Patater (changes requested):

azure-pipelines/build-and-deploy.yml
```diff
@@ -33,10 +34,6 @@ stages:
                 - script: |
                     python -m pip install --upgrade tox
                     tox -e py37
-                  displayName: "Run pytests"
```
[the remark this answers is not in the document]
rwalton-arm: Fixed.

azure-pipelines/prep-release.yml
```diff
@@ -0,0 +1,39 @@
+trigger:
+  - none
+
+pr:
+  - none
+
+schedules:
+  - cron: "00 5 * * *"
```
Patater, line 8: What will happen if there have been no changes since last release? This would happen on weekends, where we haven't merged anything since the previous night.

rwalton-arm: If there are no news files that would affect the version, the pipeline will fail and won't create a release.

Patater: Yes, I suspected the pipeline would fail. I guess we need to be okay with failures of the pipeline, or add some logic to detect change and exit the pipeline with no error. It seems best to be that failure of the pipeline should be reserved for when the pipeline isn't working as intended, rather than being nominal, so we can use failure to prompt ourselves to go fix the pipeline. The failure signal is going to be less meaningful (and perhaps mask real failures) if it is accepted as normal.

rwalton-arm: Changed the script to exit early when the version can't be bumped instead of failing.

Patater (approved):
We should be able to get a passing CI after a rebase

Patater (approved):

PR #138, opened 2020-12-01
config: Always quote option values

### LINKED ISSUE #125: mbed-tools fails to handle config options that contain quotes
### Description

From @rajkan01:

For the below mbed_lib.json config

```
"iotc-mqtt-host": { 	 
    "help": "IOTC MQTT host configuration. Defaults to mqtt.2030.ltsapis.goog host and port number 8883 if undefined", 	 
    "value": "{\"mqtt.2030.ltsapis.goog\", IOTC_MQTT_PORT}", 	 
    "macro_name": "IOTC_MQTT_HOST" 	 
 }
```

Mbedtools is generating `"-DIOTC_MQTT_HOST={"mqtt.2030.ltsapis.goog", IOTC_MQTT_PORT}"` config starts double-quotes from -D itself, and CMake prepossessing time this macro gets divided into multiple #define like below because of this define begin with double-quotes and also the value ("mqtt.2030.ltsapis.goog") with double-quote consider to be a string

```
#define IOTC_MQTT_HOST { 
#define mqtt .2030.ltsapis.goog, IOTC_MQTT_PORT} 1 
```

Could someone check this, why is the mbedtools generating macros starts with double-quotes which include `-D` and fix.

I've attached `main.ii` and `mbed_config.cmake`
[mbed_config.zip](https://github.com/ARMmbed/mbed-tools/files/5602300/mbed_config.zip)

### Issue request type

- [ ] Enhancement
- [X] Bug

Patater:
### Description

Always quote option values, being sure to escape literal quotes.
(Parenthesis don't need escaping within strings, so we also remove logic
to catch and handle the presence of parenthesis.) This prevents CMake
from getting confused when quotes are encounted within an option value
and making multiple partial list items from a single list item as
happens with the following mbed_lib.json snippet.

    "iotc-mqtt-host": {
        "help": "IOTC MQTT host configuration. Defaults to mqtt.2030.ltsapis.goog host and port number 8883 if undefined",
        "value": "{\"mqtt.2030.ltsapis.goog\", IOTC_MQTT_PORT}",
        "macro_name": "IOTC_MQTT_HOST"
    }

Which would produce the following bad CMake output (parsed as three list
items):

    "-DIOTC_MQTT_HOST={"mqtt.2030.ltsapis.goog", IOTC_MQTT_PORT}"

Now, after our changed, the output is good CMake output (parsed as one
list item):

    "-DIOTC_MQTT_HOST={\"mqtt.2030.ltsapis.goog\", IOTC_MQTT_PORT}"

Fixes #125

### Test Coverage

### DIFF
--- a/tests/build/_internal/test_cmake_file.py
+++ b/tests/build/_internal/test_cmake_file.py
@@ -2,63 +2,58 @@
 # Copyright (C) 2020 Arm Mbed. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
+import pytest
+
 from tests.build._internal.config.factories import ConfigFactory
 from mbed_tools.build._internal.cmake_file import generate_mbed_config_cmake_file, _render_mbed_config_cmake_template
 
 
+TOOLCHAIN_NAME = "gcc"
+
+
+@pytest.fixture()
+def fake_target():
+    return {
+        "labels": ["foo"],
+        "extra_labels": ["morefoo"],
+        "features": ["bar"],
+        "components": ["baz"],
+        "macros": ["macbaz"],
+        "device_has": ["stuff"],
+        "c_lib": ["c_lib"],
+        "core": ["core"],
+        "printf_lib": ["printf_lib"],
+        "supported_form_factors": ["arduino"],
+        "supported_c_libs": {TOOLCHAIN_NAME: ["ginormous"]},
+        "supported_application_profiles": ["full", "bare-metal"],
+    }
+
+
 class TestGenerateCMakeListsFile:
-    def test_correct_arguments_passed(self):
-        target = dict()
-        target["labels"] = ["foo"]
-        target["extra_labels"] = ["morefoo"]
-        target["features"] = ["bar"]
-        target["components"] = ["baz"]
-        target["macros"] = ["macbaz"]
-        target["device_has"] = ["stuff"]
-        target["c_lib"] = ["c_lib"]
-        target["core"] = ["core"]
-        target["printf_lib"] = ["printf_lib"]
-        target["supported_form_factors"] = ["arduino"]
+    def test_correct_arguments_passed(self, fake_target):
         config = ConfigFactory()
         mbed_target = "K64F"
-        toolchain_name = "GCC"
-        target["supported_c_libs"] = {toolchain_name.lower(): ["small", "std"]}
-        target["supported_application_profiles"] = ["full", "bare-metal"]
 
-        result = generate_mbed_config_cmake_file(mbed_target, target, config, toolchain_name)
+        result = generate_mbed_config_cmake_file(mbed_target, fake_target, config, TOOLCHAIN_NAME)
 
-        assert result == _render_mbed_config_cmake_template(target, config, toolchain_name, mbed_target,)
+        assert result == _render_mbed_config_cmake_template(fake_target, config, TOOLCHAIN_NAME, mbed_target,)
 
 
 class TestRendersCMakeListsFile:
-    def test_returns_rendered_content(self):
-        target = dict()
-        target["labels"] = ["foo"]
-        target["extra_labels"] = ["morefoo"]
-        target["features"] = ["bar"]
-        target["components"] = ["baz"]
-        target["macros"] = ["macbaz"]
-        target["device_has"] = ["stuff"]
-        target["core"] = ["core"]
-        target["c_lib"] = ["c_lib"]
-        target["printf_lib"] = ["printf_lib"]
-        target["supported_form_factors"] = ["arduino"]
+    def test_returns_rendered_content(self, fake_target):
         config = ConfigFactory()
-        toolchain_name = "baz"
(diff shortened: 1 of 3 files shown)


tests/build/_internal/test_cmake_file.py
```diff
@@ -2,13 +2,13 @@
 # Copyright (C) 2020 Arm Mbed. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-from unittest import TestCase
+import pytest
```
rwalton-arm, line 5: We shouldn't need to import `pytest` as we're not using anything from that module in commit  "tests: Refactor CMake config test to use pytest". The pytest runner doesn't require this import statement to pick up the test.

tests/build/_internal/test_cmake_file.py
```diff
@@ -7,60 +7,51 @@
 from tests.build._internal.config.factories import ConfigFactory
 from mbed_tools.build._internal.cmake_file import generate_mbed_config_cmake_file, _render_mbed_config_cmake_template
 
+@pytest.fixture()
+def fake_target():
+    target = dict()
```
rwalton-arm, line 14: Why not just set the dictionary key-value pairs when it's constructed?

```
def fake_target():
    yield {"labels": ["foo"], ...}
```

tests/build/_internal/test_cmake_file.py
```diff
@@ -7,60 +7,51 @@
...
-        target["c_lib"] = ["c_lib"]
-        target["printf_lib"] = ["printf_lib"]
-        target["supported_form_factors"] = ["arduino"]
+    def test_returns_rendered_content(self, fake_target):
         config = ConfigFactory()
-        toolchain_name = "baz"
-        target["supported_c_libs"] = {toolchain_name.lower(): ["small", "std"]}
-        target["supported_application_profiles"] = ["full", "bare-metal"]
-        result = _render_mbed_config_cmake_template(target, config, toolchain_name, "target_name")
+        toolchain_name = "loo"
```
rwalton-arm, line 44: Why did we change the name of this?

tests/build/_internal/test_cmake_file.py
Patater: I'll try it and see if the line wrapping still looks ok.

tests/build/_internal/test_cmake_file.py
Patater: Yes, I only need pytest later when I start using a fixture. I'll move it commits.

tests/build/_internal/test_cmake_file.py
Patater: There wasn't a name that both tests used, and it is required that this name match what's in the fixture for the fake_target. Any better ideas on how to match names? I don't like repeating the name.

tests/build/_internal/test_cmake_file.py
```diff
@@ -7,60 +7,51 @@
...
+    target["components"] = ["baz"]
+    target["macros"] = ["macbaz"]
+    target["device_has"] = ["stuff"]
+    target["c_lib"] = ["c_lib"]
+    target["core"] = ["core"]
+    target["printf_lib"] = ["printf_lib"]
+    target["supported_form_factors"] = ["arduino"]
+    target["supported_c_libs"] = {"loo": ["ginormous"]}
+    target["supported_application_profiles"] = ["full", "bare-metal"]
+    yield target
```
rwalton-arm, line 27: We can just use `return` here as we don't need to pass control back to the fixture after the caller has finished executing in this case. We would use `yield` if we wanted the fixture to act as a context manager, or we needed to perform some tear-down step after the test has completed.

rwalton-arm (changes requested):

tests/build/_internal/test_cmake_file.py
rwalton-arm: Could we define a `toolchain_name` fixture that just returns the name as a string?

Patater: I made a global. :/

tests/build/_internal/test_cmake_file.py
```diff
@@ -2,63 +2,58 @@
 # Copyright (C) 2020 Arm Mbed. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
+import pytest
+
 from tests.build._internal.config.factories import ConfigFactory
 from mbed_tools.build._internal.cmake_file import generate_mbed_config_cmake_file, _render_mbed_config_cmake_template
 
 
+toolchain_name = "gcc"
```
rwalton-arm, line 12: We should use all caps with underscores for global constants.

rwalton-arm (changes requested):

tests/build/_internal/test_cmake_file.py
Patater: OK

rwalton-arm (approved):

PR #139, opened 2020-12-01
Add regression tests for config processing

rwalton-arm:
### Description

Add regression tests at the public interface level so we can rework the
config processing code and have some confidence we haven't broken anything.

### Test Coverage

### DIFF
--- a/tests/build/test_generate_config.py
+++ b/tests/build/test_generate_config.py
@@ -2,31 +2,396 @@
 # Copyright (C) 2020 Arm Mbed. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-from unittest import TestCase, mock
+import json
 
+import pytest
+
+from mbed_tools.project import MbedProgram
 from mbed_tools.build import generate_config
+from mbed_tools.lib.exceptions import ToolsError
 
 
-class TestGenerateConfig(TestCase):
-    @mock.patch("mbed_tools.build.config.MbedProgram")
-    @mock.patch("mbed_tools.build.config.get_target_by_name")
-    @mock.patch("mbed_tools.build.config.generate_mbed_config_cmake_file")
-    @mock.patch("mbed_tools.build.config.assemble_config")
-    @mock.patch("mbed_tools.build.config.write_file")
-    def test_collaborators_called_with_corrrect_arguments(
-        self, write_file, assemble_config, gen_config_cmake, get_target_by_name, mbed_program
-    ):
-        program = mbed_program.from_existing()
+TARGETS = ["K64F", "NUCLEO_F401RE"]
 
-        generate_config("K64F", "GCC_ARM", program)
+TARGET_DATA = {
+    "bootloader_supported": True,
+    "c_lib": "std",
+    "components": ["FLASHIAP", "SD"],
+    "config": {
+        "xip-enable": {
+            "help": "Enable Execute In Place (XIP) on this "
+            "target. Value is only significant if the "
+            "board has executable external storage such "
+            "as QSPIF. If this is enabled, customize "
+            "the linker file to choose what text "
+            "segments are placed on external storage",
+            "value": False,
+        }
+    },
+    "core": "Cortex-M4F",
+    "default_toolchain": "ARM",
+    "detect_code": ["0240"],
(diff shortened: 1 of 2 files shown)


tests/build/test_generate_config.py
```diff
@@ -2,31 +2,294 @@
...
+
+def test_macros_from_lib_collected(program):
+    macros = ["NS_USE_EXTERNAL_MBED_TLS"]
+    create_mbed_lib_json(program.mbed_os.root / "connectivity" / "mbed_lib.json", "nanostack", macros=macros)
+
+    generate_config("K64F", "GCC_ARM", program)
+
+    config_text = program.files.cmake_config_file.read_text()
+
+    for macro in macros:
```
Patater, line 147: Do you want to test for more than one macro? Probably in a separate test.

tests/build/test_generate_config.py
```diff
@@ -2,31 +2,294 @@
...
+        assert f"FEATURE_{feature}=1" in config_text
+
+    for device in TARGET_DATA["device_has"]:
+        assert f"DEVICE_{device}=1" in config_text
+
+    for form_factor in TARGET_DATA["supported_form_factors"]:
+        assert f"TARGET_FF_{form_factor}" in config_text
+
+
+def test_overrides_lib_config_param_from_app(matching_target_and_filter, program):
```
Patater, line 178: A few more test ideas:
- Does a lib_config override a target config?
- Does a lib_config override another, previously parsed, lib_config?
- Are separate settings from multiple lib_configs all included?

tests/build/test_generate_config.py
```diff
@@ -2,31 +2,294 @@
...
+    config_text = program.files.cmake_config_file.read_text()
+
+    assert "MBED_CONF_TARGET_XIP_ENABLE=1" in config_text
+
+
+@pytest.mark.parametrize(
+    "config_param, config_param_value, expected_output",
+    [
+        ("target.c_lib", "super", 'MBED_C_LIB "super"'),
+        ("target.printf_lib", "maximal-printf", 'MBED_PRINTF_LIB "maximal-printf"'),
```
Patater, line 207: Any other `target.` attributes we should test here?

tests/build/test_generate_config.py
```diff
@@ -2,31 +2,294 @@
...
+    assert "MBED_CONF_PLATFORM_STDIO_BAUD_RATE=115200" in config_text
+
+
+@pytest.mark.parametrize(
+    "config_param, config_param_value, expected_output",
+    [
+        ("target.macros_add", ["ENABLE_BOBBINS"], TARGET_DATA["macros"] + ["ENABLE_BOBBINS"]),
+        ("target.extra_labels_add", ["EXTRA_LABEL_BOBBINS"], TARGET_DATA["extra_labels"] + ["EXTRA_LABEL_BOBBINS"]),
+        ("target.features_add", ["FEATURE_BOBBINS"], TARGET_DATA["features"] + ["FEATURE_BOBBINS"]),
+        ("target.components_add", ["COMPONENT_BOBBINS"], TARGET_DATA["components"] + ["COMPONENT_BOBBINS"]),
```
Patater, line 245: mmm, bobbins.

tests/build/test_generate_config.py
rwalton-arm: These are currently the only two that work. I'll add more in as xfails, raise an issue and fix them when I rework this code.

tests/build/test_generate_config.py
rwalton-arm: We have an issue where lib configs can override other lib configs, the order of overrides being undefined if this is attempted in multiple libs. The old tools raised an error in that situation. I will add a test to check we raise an error and mark it as xfail until the bug is fixed (I will also raise a defect to track it).

Patater: Thanks for raising https://github.com/ARMmbed/mbed-tools/issues/155

tests/build/test_generate_config.py
Patater: Thanks for raising https://github.com/ARMmbed/mbed-tools/issues/156

Patater (approved):

Issue #127, opened 2020-11-26
Lack of handling of "requires" in mbed_app.json


LDong-Arm:
### Description

The Mbed configuration system has a `"requires"` field (see [example](https://github.com/ARMmbed/mbed-os-example-blinky-baremetal/blob/master/mbed_app.json)) to restrict compilation to certain libraries instead of the whole Mbed OS. This also affects _what macros get defined_, i.e. macros are only generated from `mbed_lib.json` of libraries we build.

The current version of mbed-tools defines macros from all libraries, regardless of `"defines"`. This causes issues (see discussion in https://github.com/ARMmbed/mbed-os/pull/13934) when we need to define a macro only when the library is enabled.

Note: If this issue is resolved, we should _not_ need https://github.com/ARMmbed/mbed-tools/pull/78 anymore. But there's a more fundamental issue: the `"requires": [...]` field duplicates the purpose CMake dependency lists:
```
target_link_libraries(${APP_TARGET}
    PRIVATE
        mbed-os
        mbed-netsocket
)
```
Such duplication means users would list dependencies twice, once for CMake, once in "requires". How could we handle this?

### Issue request type

- [ ] Enhancement
- [x] Bug

LDong-Arm:
@Patater @hugueskamba

PR #158, opened 2020-12-14
config: Add support for `requires` config option

urutva:
### Description

Support for a config option `requires` in `mbed_app.json` is added
which enables application to choose required mbed libraries. if
`requires` is mentioned in `mbed_app.json` then only those mbed
libraries and other libraries those mbed libraries depend on will be
selected. Other mbed libraries are ignored.

Fixes https://github.com/ARMmbed/mbed-tools/issues/127.

### Test Coverage

### DIFF
--- a/src/mbed_tools/build/_internal/config/assemble_build_config.py
+++ b/src/mbed_tools/build/_internal/config/assemble_build_config.py
@@ -9,7 +9,8 @@ from typing import Iterable, Optional
 from mbed_tools.build._internal.config.config import Config
 from mbed_tools.build._internal.config.cumulative_data import CumulativeData
 from mbed_tools.build._internal.config.source import Source
-from mbed_tools.build._internal.find_files import LabelFilter, filter_files, find_files
+from mbed_tools.lib.json_helpers import decode_json_file
+from mbed_tools.build._internal.find_files import LabelFilter, RequiresFilter, filter_files, find_files
 
 
 def assemble_config(target_attributes: dict, mbed_program_directory: Path, mbed_app_file: Optional[Path]) -> Config:
@@ -27,13 +28,23 @@ def _assemble_config_from_sources_and_lib_files(
     target_attributes: dict, mbed_lib_files: Iterable[Path], mbed_app_file: Optional[Path] = None
 ) -> Config:
     previous_cumulative_data = None
+    requires = list()
     target_source = Source.from_target(target_attributes)
     current_cumulative_data = CumulativeData.from_sources([target_source])
+    if mbed_app_file:
+        app_data = decode_json_file(mbed_app_file)
+        requires = app_data["requires"] if "requires" in app_data else []
+
     while previous_cumulative_data != current_cumulative_data:
         current_labels = current_cumulative_data.labels | current_cumulative_data.extra_labels
         filtered_files = _filter_files(
-            mbed_lib_files, current_labels, current_cumulative_data.features, current_cumulative_data.components
+            mbed_lib_files,
+            current_labels,
+            current_cumulative_data.features,
+            current_cumulative_data.components,
+            requires,
         )
+
         mbed_lib_sources = [Source.from_mbed_lib(file, current_labels) for file in filtered_files]
         all_sources = [target_source] + mbed_lib_sources
         if mbed_app_file:
@@ -48,14 +59,24 @@ def _assemble_config_from_sources_and_lib_files(
 
 
 def _filter_files(
-    files: Iterable[Path], labels: Iterable[str], features: Iterable[str], components: Iterable[str]
+    files: Iterable[Path],
+    labels: Iterable[str],
+    features: Iterable[str],
+    components: Iterable[str],
+    requires: Iterable[str],
 ) -> Iterable[Path]:
     filters = (
         LabelFilter("TARGET", labels),
         LabelFilter("FEATURE", features),
         LabelFilter("COMPONENT", components),
     )
-    return filter_files(files, filters)
+    filtered_files = filter_files(files, filters)
+
+    if not requires:
+        return filtered_files
+    else:
+        requires_filter = RequiresFilter(requires)
+        return requires_filter(filtered_files)
 
 
 def _update_target_attributes(target_attributes: dict, cumulative_data: CumulativeData) -> None:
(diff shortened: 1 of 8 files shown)


tests/build/test_generate_config.py
```diff
@@ -395,3 +395,35 @@ def test_settings_from_multiple_libs_included(matching_target_and_filter, progra
 
     assert "MBED_CONF_PLATFORM_STDIO_BAUD_RATE=9600" in config_text
     assert "MBED_LFS_READ_SIZE=64" in config_text
+
+
+def test_requires_config_option(program):
```
rwalton-arm, line 400: Was this test failing before we added the `requires` handling? Do we also need a test to ensure we're filtering out things we don't `require`?

src/mbed_tools/build/_internal/config/assemble_build_config.py
```diff
@@ -47,6 +60,20 @@ def _assemble_config_from_sources_and_lib_files(
...
+
+    The 'requires' config option in mbed_app.json can specify list of mbed
+    libraries (mbed_lib.json) that application requires. Apply 'requires'
+    filter to remove mbed_lib.json files not required by application.
+    """
+    for required_mbed_lib in requires:
+        for mbed_lib_path in files:
+            if required_mbed_lib == _get_mbed_lib_name(mbed_lib_path):
+                requires_filtered_files.add(mbed_lib_path)
+                _apply_requires_filter(requires_filtered_files, files, _get_mbed_lib_requires(mbed_lib_path))
```
rwalton-arm, line 74: Is it possible to just extend `_filter_files` with another filter type, instead of introducing this extra filtering mechanism?

rwalton-arm (changes requested):

src/mbed_tools/build/_internal/config/assemble_build_config.py
```diff
@@ -67,3 +94,23 @@ def _update_target_attributes(target_attributes: dict, cumulative_data: Cumulati
     target_attributes["macros"] = cumulative_data.macros
     target_attributes["c_lib"] = cumulative_data.c_lib
     target_attributes["printf_lib"] = cumulative_data.printf_lib
+
+
+def _get_mbed_lib_requires(mbed_lib_path: Path) -> Any:
+    """Get list of mbed libraries required by appplication."""
+    json_content = _get_json_contents(mbed_lib_path)
```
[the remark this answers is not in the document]
urutva: Ah good point. Fixed.

src/mbed_tools/build/_internal/config/assemble_build_config.py
```diff
@@ -67,3 +94,23 @@ def _update_target_attributes(target_attributes: dict, cumulative_data: Cumulati
...
+    return json_content["requires"] if "requires" in json_content else []
+
+
+def _get_mbed_lib_name(mbed_lib_path: Path) -> Any:
+    """Get mbed library name."""
+    return _get_json_contents(mbed_lib_path)["name"]
+
+
+def _get_json_contents(path: Path) -> Any:
```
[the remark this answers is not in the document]
urutva: Fixed.

tests/build/test_generate_config.py
urutva: Yes, this test will fail without `requires` handling.

This test combines both checking for config options that applications requires and don't require.
```
assert "MBED_CONF_PLATFORM_STDIO_BAUD_RATE=9600" in config_text
    assert "MBED_LFS_READ_SIZE=64" not in config_text
```
Should I split them to different tests? Most of the contents of the test will remain the same. So I decided to put them together.

rwalton-arm: OK. I missed the `not` in the second assertion when I was looking over the code. Keeping it all in one test is fine.

0xc0170:
>  then only those mbed
libraries and other libraries those mbed libraries depend on will be
selected. Other mbed libraries are ignored.

How does this work? I would like to understand the implications for CMake. What does "selected" mean in this context?

src/mbed_tools/build/_internal/config/assemble_build_config.py
urutva: Updated `_filter_files` with requires filter type and moved `_apply_requires_filter` requires filter.

urutva:
> > then only those mbed
> > libraries and other libraries those mbed libraries depend on will be
> > selected. Other mbed libraries are ignored.
> 
> How does this work? I would like to understand the implications for CMake. What does "selected" mean in this context?

@0xc0170 This PR adds support for `requires` config option in `mbed_app.json`. The issue is described in https://github.com/ARMmbed/mbed-tools/issues/127.

The behaviour of new tools (Mbed CLI 2) is same that of old tool (Mbed CLI 1) with respect of handling of `requires`. That is, if `requires` is mentioned in `mbed_app.json` then only those libraries (mbed_lib.json) and the other libraries they depend on will be used to generate the configuration options (`.mbedbuild/mbed_config.cmake`). All other libraries (mbed_lib.json) are ignored.

Unfortunately, as described in https://github.com/ARMmbed/mbed-tools/issues/127, `requires` in `mbed_app.json` is duplication of `target_link_libraries` in `CMakeList.txt`. But this is unavoidable as we don't have any mechanism to feed the information from CMake to mbed-tools.

src/mbed_tools/build/_internal/find_files.py
```diff
@@ -66,6 +68,56 @@ def filter_files(files: Iterable[Path], filters: Iterable[Callable]) -> Iterable
...
+    def __init__(self, requires: Iterable[str]):
+        """Initialise the filter attributes.
+
+        Args:
+            requires: List of required mbed libraries.
+        """
+        self._requires = requires
+
+    def __call__(self, files: Iterable[Path]) -> Iterable[Path]:
+        """Return True if given path contains only allowed labels - should not be filtered out."""
```
rwalton-arm, line 88: This docstring seems misleading. We're actually returning an `Iterable[Path]` here. Should we cast it to a `bool` if we only care about `True`/`False`?

rwalton-arm (changes requested):

src/mbed_tools/build/_internal/find_files.py
urutva: Fixed.

rwalton-arm (approved):

PR #168, opened 2021-01-21
compile: Update command line option

urutva:
### Description

In Mbed CLI 1, `--profile` was used to identify type of build. Change
`--build-type` to `--profile` to match Mbed CLI 1 behaviour.

### Test Coverage

### DIFF
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -23,7 +23,7 @@ from mbed_tools.sterm import terminal
     help="The toolchain you are using to build your app.",
 )
 @click.option("-m", "--mbed-target", help="A build target for an Mbed-enabled device, e.g. K64F.")
-@click.option("-b", "--build-type", default="develop", help="The build type (release, develop or debug).")
+@click.option("-b", "--profile", default="develop", help="The build type (release, develop or debug).")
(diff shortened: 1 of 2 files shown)


src/mbed_tools/cli/build.py
```diff
@@ -23,7 +23,7 @@
     help="The toolchain you are using to build your app.",
 )
 @click.option("-m", "--mbed-target", help="A build target for an Mbed-enabled device, e.g. K64F.")
-@click.option("-b", "--build-type", default="develop", help="The build type (release, develop or debug).")
+@click.option("--profile", default="develop", help="The build type (release, develop or debug).")
```
Patater, line 26: Should we keep "-b" for this?

urutva: Usually, first letter is used as an alias. But we are already using `-p` for `--program-path`. So I thought of removing `-b`.

urutva: Since backward compatibility can be achieved through `-b` option, should I change the `major` to `bugfix`?

Patater (approved):

src/mbed_tools/cli/build.py
Patater: The full option changed, which breaks backwards compatibility still.

PR #170, opened 2021-01-26
Add some mbed-os dependencies as our own

Patater:
### Description

Add prettytable and intehex as installation dependencies to avoid new
installs from being complicated by the fact that one must go into
mbed-os to install requirements.txt. Yes, these aren't real requirements
of mbed-tools. Yes, this is a bit of an architecture violation. However,
until we can address the problem in a more suitable way, this should
help avoid a common usability impediment.

### Test Coverage

- [X]  This change is covered by existing or additional automated tests.


setup.py
```diff
@@ -49,6 +50,10 @@
         "typing-extensions",
         "Jinja2",
         "pyserial",
+        # Mbed OS 6.6 dependencies
+        "prettytable==0.7.2",
```
rwalton-arm, line 54: Can we get away without pinning the versions? I can see dependency hell in our future.

Patater: We'll have some dependency hell until we have an Mbed Installer, as adding these dependencies now means that the mbed-os version and the mbed-tools version are coupled: you can no longer build any version of Mbed OS 6.5 and up with any mbed-tools version.

Patater: I've removed the versions from here. Hopefully mbed-os still works.

rwalton-arm: I think we need some more explanation about why we're adding these dependencies in the comment. Also possibly a "TODO: remove this hack when we have an Mbed installer"

Patater: Commit message expanded with "what if we didn't fix this" description and a link to an issue tracking removal of this usability hack. TODO also added to setup.py itself.

pyproject.toml
```diff
@@ -77,8 +77,3 @@ showcontent = false
 
 [tool.black]
 line-length = 120
```
rwalton-arm, line 79: We should also be able to remove the following section now we've eliminated the SPDX report: https://github.com/ARMmbed/mbed-tools/blob/d3c251d0d43b2aba272aff73c6e152adfef86380/pyproject.toml#L25

Patater: Removed

rwalton-arm (approved):

PR #175, opened 2021-02-02
Add build subdirectories for multiple targets/toolchains

### LINKED ISSUE #141: Make a different cmake_build tree for ARM and GCC and for each target
### Description

Hi

With previous system build, ARM and GCC built files are in a different tree.
I could then verify development and applications with both toolchain in //

Now it is no more possible...

In previous build, development and debug profiles were also different.

And I think seperating os build from baremetal build would be also appreciated :-)

Thx

### Issue request type

- [x] Enhancement
- [ ] Bug

wernerlewis:
### Description

Creates new subdirectories in cmake_build for each target/profile/toolchain combination, and writes build tree into these. This will be in `cmake_build/TARGET/PROFILE/TOOLCHAIN`

With these changes multiple target-toolchain combinations can be compiled for without manually moving files or clearing parts of the build tree each time. To ensure clarity with use of multiple targets and toolchains, both will be made to be required arguments for `compile`.

Fixes #141

### Test Coverage

### DIFF
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -87,10 +87,11 @@ def build(
        baudrate: Change the serial baud rate (ignored unless --sterm is also given).
     """
     _validate_target_and_toolchain_args(mbed_target, toolchain)
+    cmake_build_subdir = pathlib.Path(mbed_target.upper(), profile.lower(), toolchain.upper())
     if mbed_os_path is None:
-        program = MbedProgram.from_existing(pathlib.Path(program_path))
+        program = MbedProgram.from_existing(pathlib.Path(program_path), cmake_build_subdir)
     else:
-        program = MbedProgram.from_existing(pathlib.Path(program_path), pathlib.Path(mbed_os_path))
+        program = MbedProgram.from_existing(pathlib.Path(program_path), cmake_build_subdir, pathlib.Path(mbed_os_path))
     build_tree = program.files.cmake_build_dir
     if clean and build_tree.exists():
         shutil.rmtree(build_tree)
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -91,23 +91,15 @@ def build(
         program = MbedProgram.from_existing(pathlib.Path(program_path))
     else:
         program = MbedProgram.from_existing(pathlib.Path(program_path), pathlib.Path(mbed_os_path))
-    mbed_config_file = program.files.cmake_config_file
     build_tree = program.files.cmake_build_dir
     if clean and build_tree.exists():
         shutil.rmtree(build_tree)
-    if any(
-        [
-            not mbed_config_file.exists(),
-            not build_tree.exists(),
-            mbed_os_path is not None,
-            custom_targets_json is not None,
-        ]
-    ):
-        click.echo("Configuring project and generating build system...")
-        if custom_targets_json is not None:
-            program.files.custom_targets_json = pathlib.Path(custom_targets_json)
-        generate_config(mbed_target.upper(), toolchain, program)
-        generate_build_system(program.root, build_tree, profile)
+
+    click.echo("Configuring project and generating build system...")
+    if custom_targets_json is not None:
+        program.files.custom_targets_json = pathlib.Path(custom_targets_json)
+    generate_config(mbed_target.upper(), toolchain, program)
+    generate_build_system(program.root, build_tree, profile)
 
     click.echo("Building Mbed project...")
     build_project(build_tree)
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -86,6 +86,7 @@ def build(
        sterm: Open a serial terminal to the connected target.
        baudrate: Change the serial baud rate (ignored unless --sterm is also given).
     """
+    _validate_target_and_toolchain_args(mbed_target, toolchain)
     if mbed_os_path is None:
         program = MbedProgram.from_existing(pathlib.Path(program_path))
     else:
@@ -94,13 +95,17 @@ def build(
     build_tree = program.files.cmake_build_dir
     if clean and build_tree.exists():
         shutil.rmtree(build_tree)
-
-    if any([not mbed_config_file.exists(), not build_tree.exists(), mbed_target, toolchain]):
+    if any(
+        [
+            not mbed_config_file.exists(),
+            not build_tree.exists(),
+            mbed_os_path is not None,
+            custom_targets_json is not None,
+        ]
+    ):
         click.echo("Configuring project and generating build system...")
-        _validate_target_and_toolchain_args(mbed_target, toolchain)
         if custom_targets_json is not None:
             program.files.custom_targets_json = pathlib.Path(custom_targets_json)
-
         generate_config(mbed_target.upper(), toolchain, program)
         generate_build_system(program.root, build_tree, profile)
 
@@ -127,4 +132,6 @@ def build(
 
 def _validate_target_and_toolchain_args(target: str, toolchain: str) -> None:
     if not all([toolchain, target]):
-        raise click.UsageError("--toolchain and --mbed-target arguments are required when generating Mbed config!")
+        raise click.UsageError(
+            "Both --toolchain and --mbed-target arguments are required when using the compile subcommand."
+        )
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -75,7 +75,7 @@ def build(
 
     Args:
        program_path: Path to the Mbed project.
-       mbed_os_path: the path to the local Mbed OS directory
+       mbed_os_path: The path to the local Mbed OS directory.
        profile: The Mbed build profile (debug, develop or release).
        custom_targets_json: Path to custom_targets.json.
        toolchain: The toolchain to use for the build.
--- a/src/mbed_tools/cli/configure.py
+++ b/src/mbed_tools/cli/configure.py
@@ -48,10 +48,11 @@ def configure(toolchain: str, mbed_target: str, program_path: str, mbed_os_path:
         program_path: the path to the local Mbed program
         mbed_os_path: the path to the local Mbed OS directory
     """
+    cmake_build_subdir = pathlib.Path(mbed_target.upper(), "develop", toolchain.upper())
     if mbed_os_path is None:
-        program = MbedProgram.from_existing(pathlib.Path(program_path))
+        program = MbedProgram.from_existing(pathlib.Path(program_path), cmake_build_subdir)
     else:
-        program = MbedProgram.from_existing(pathlib.Path(program_path), pathlib.Path(mbed_os_path))
+        program = MbedProgram.from_existing(pathlib.Path(program_path), cmake_build_subdir, pathlib.Path(mbed_os_path))
     mbed_target = mbed_target.upper()
     output_path = generate_config(mbed_target, toolchain, program)
     click.echo(f"mbed_config.cmake has been generated and written to '{str(output_path.resolve())}'")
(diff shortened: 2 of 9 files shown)


0xc0170:
Can you describe how this changes the output in the build directory (without going into the code changes to find out). 

What should be the structure in the build directory and how this matches the Mbed CLI 1 behavior (the same ,different ?).

wernerlewis:
@0xc0170 
The build files will now be written in: `cmake_build/TARGET/PROFILE/TOOLCHAIN/`
Config files will be written in:  `cmake_build/TARGET/config/TOOLCHAIN/`

Whenever a new target, toolchain, profile combination is used it will create the subdirectories as needed, and the output will be written into there. It's similar behavior to Mbed CLI 1 with the additional profile subdirectory (for develop, build, release) .

Patater:
> @0xc0170
> The build files will now be written in: `cmake_build/TARGET/PROFILE/TOOLCHAIN/`
> Config files will be written in: `cmake_build/TARGET/config/TOOLCHAIN/`
> 
> Whenever a new target, toolchain, profile combination is used it will create the subdirectories as needed, and the output will be written into there. It's similar behavior to Mbed CLI 1 with the additional profile subdirectory (for develop, build, release) .

Could you put this into the commit message, please?

src/mbed_tools/build/config.py
```diff
@@ -12,6 +12,10 @@
 from mbed_tools.build._internal.write_files import write_file
 
 
+DEFAULT_CONFIG_DIR = ".mbedbuild"
```
Patater, line 15: In the commit description for "config: Write config file to target subdirectory "...

> Writes generated config files to the specific target/develop/toolchain/
subdirectory as well as .mbedbuild.

Could you please explain why it is good to put the generated config files into both places? It might be good to remark explicitly on why we aren't putting the generated config files to only the specific subdirectory.

src/mbed_tools/project/_internal/project_data.py
```diff
@@ -97,12 +98,24 @@ def from_new(cls, root_path: Path) -> "MbedProgramFiles":
...
 
         Args:
             root_path: The path containing the MbedProgramFiles.
+            target_name: Name of the Mbed target in use.
+            profile: Current profile for build.
+            toolchain: Name of the toolchain in use.
         """
+        if target_name is None or toolchain is None:
+            default_cmake_config = root_path / CMAKE_CONFIG_FILE_PATH
+            target_name, toolchain = _prev_target_toolchain(default_cmake_config)
```
rwalton-arm, line 119: I think we should just make users pass the target and toolchain on the command line instead of reading values from mbed_config.cmake.

src/mbed_tools/cli/build.py
```diff
@@ -82,9 +82,18 @@ def build(
...
+        program = MbedProgram.from_existing(
+            pathlib.Path(program_path),
+            pathlib.Path(mbed_os_path),
+            target_name=mbed_target,
+            profile=profile,
+            toolchain=toolchain,
+        )
+
     mbed_config_file = program.files.cmake_config_file
     build_tree = program.files.cmake_build_dir
```
rwalton-arm, line 98: Instead of passing target, toolchain and the kitchen sink to `MbedProgram` could we just construct the full `build_dir` path here? I think it would simplify this code a lot.

rwalton-arm (changes requested):

Patater (changes requested):
I don't see any documentation updates in this PR. Could you confirm you've looked over the README and other documentation to determine if we need to update anything? For docs, we might need a follow-on PR to [the Mbed CLI 2 docs in mbed-os-5-docs](https://github.com/ARMmbed/mbed-os-5-docs/tree/development/docs/tools/mbed_cli_2), given that the code and docs live in separate repos.

wernerlewis:
> I don't see any documentation updates in this PR. Could you confirm you've looked over the README and other documentation to determine if we need to update anything? For docs, we might need a follow-on PR to [the Mbed CLI 2 docs in mbed-os-5-docs](https://github.com/ARMmbed/mbed-os-5-docs/tree/development/docs/tools/mbed_cli_2), given that the code and docs live in separate repos.

I checked documentation and I don't think there are any changes needed in this repo, but it will need a PR in the other repo though due to the change in behavior.

src/mbed_tools/project/_internal/project_data.py
```diff
@@ -120,7 +122,7 @@ def from_existing(cls, root_path: Path) -> "MbedProgramFiles":
             mbed_os_ref=mbed_os_file,
             cmakelists_file=cmakelists_file,
             cmake_config_file=root_path / CMAKE_CONFIG_FILE_PATH,
-            cmake_build_dir=root_path / CMAKE_BUILD_DIR,
+            cmake_build_dir=root_path / CMAKE_BUILD_DIR / build_subdir,
```
Patater, line 130: Isn't the CMAKE dir the childmost directory? This line makes it sound like the cmake directory has the profile-specific folders under it. Probably `CMAKE_BUILD_DIR` is just a misleading name. Consider updating it to something like `BUILD_DIR`, to avoid implying that the cmake build directory is that directory.

news/20210203121128.feature
```diff
@@ -0,0 +1 @@
+Add mbed_config.cmake to cmake_build subdirectories for each target and toolchain combination.
```
Patater, line 1: What does this enable the user to do?

Patater (changes requested):

src/mbed_tools/cli/build.py
```diff
@@ -81,20 +81,18 @@ def build(
...
     if clean and build_tree.exists():
         shutil.rmtree(build_tree)
 
-    if any([not mbed_config_file.exists(), not build_tree.exists(), mbed_target, toolchain]):
-        click.echo("Configuring project and generating build system...")
-        _validate_target_and_toolchain_args(mbed_target, toolchain)
-        generate_config(mbed_target.upper(), toolchain, program)
-        generate_build_system(program.root, build_tree, profile)
+    click.echo("Configuring project and generating build system...")
```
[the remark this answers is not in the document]
wernerlewis: I'll revise this. Looking at it again it makes much more sense to drop the conditions where it will always generate with a new target and toolchain, rather than dropping it altogether, when supporting multiple targets.

src/mbed_tools/project/_internal/project_data.py
```diff
@@ -114,13 +115,21 @@ def from_existing(cls, root_path: Path) -> "MbedProgramFiles":
         cmakelists_file = root_path / CMAKELISTS_FILE_NAME
         if not cmakelists_file.exists():
             logger.warning("No CMakeLists.txt found in the program root.")
+        if build_subdir is None:
```
Patater, line 118: Are we allowing build_subdir to be None? What does it mean when build_subdir is None? I thought we wanted to always write to a target-specific subdirectory. This looks like it it making a new "None" folder sometimes.

wernerlewis: I added None as an option so that tests would be able to check that from_existing and from_new will produce the same MbedProgramFiles, since from_new doesn't take target and toolchain arguments (and so doesn't point to a configuration specific build_directory). It shouldn't ever create the "None" subdirectory as the build_subdir argument is always used outside of tests, but I can remove the / "None" to reduce confusion.

Patater: Maybe there is another way to test this, perhaps with mocking. Then we wouldn't need to add "test only" code to our production Python files (just the test ones).

tests/project/test_mbed_program.py
```diff
@@ -10,17 +10,20 @@
 from mbed_tools.project import MbedProgram
 from mbed_tools.project.exceptions import ExistingProgram, ProgramNotFound, MbedOSNotFound
 from mbed_tools.project.mbed_program import _find_program_root, parse_url
-from mbed_tools.project._internal.project_data import MbedProgramFiles
+from mbed_tools.project._internal.project_data import CMAKE_CONFIG_FILE_NAME, MbedProgramFiles
 from tests.project.factories import make_mbed_program_files, make_mbed_os_files, patchfs
 
 
 DEFAULT_BUILD_SUBDIR = pathlib.Path("K64F", "develop", "GCC_ARM")
+DEFAULT_CONFIG_SUBDIR = pathlib.Path("K64F", "config", "GCC_ARM")
```
Patater, line 18: Shouldn't the default config subdir be `.mbedbuild` or `.`? Which path relative to the project does CMake read mbed_config.cmake from usually?

src/mbed_tools/cli/build.py
```diff
@@ -94,13 +95,17 @@ def build(
...
         shutil.rmtree(build_tree)
-
-    if any([not mbed_config_file.exists(), not build_tree.exists(), mbed_target, toolchain]):
+    if any(
+        [
+            not mbed_config_file.exists(),
+            not build_tree.exists(),
+            mbed_os_path is not None,
+            custom_targets_json is not None,
+        ]
```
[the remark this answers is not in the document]
wernerlewis: I had thought if the user has a custom_target_json arg then they would want to configure, as that won't be used otherwise, but maybe it would be better if they had to explicitly choose to reconfigure.

tests/project/test_mbed_program.py
wernerlewis: CMake does use the config from `.mbedbuild`, while this would be the path to the retained config for the current target and toolchain. It would be clearer if I were to revert cmake_config_file to `.mbedbuild/mbed_config.cmake` and then just make the backup during `configure`, as the retained config isn't used.

src/mbed_tools/cli/build.py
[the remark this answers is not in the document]
rwalton-arm: Why don't we just always reconfigure? CMake should be smart enough to figure out if we need an incremental build or not.

tests/project/test_mbed_program.py
rwalton-arm: Why do we need to "retain the config for the current target and toolchain" if we make the user always pass target and toolchain? This backup stuff sounds unnecessary to me.

src/mbed_tools/cli/build.py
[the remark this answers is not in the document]
Patater: Looks we do indeed want to unconditionally configure, even when the build tree exists already. Apologies for the back and forth on this. We should make this change in a separate commit and explain the reasoning in the commit message.

tests/project/test_mbed_program.py
wernerlewis: Just to allow the user to check the previously generated config that is associated with each build, without needing to generate each one again, if that isn't necessary I can drop it.

Patater (approved):
LGTM

rwalton-arm (approved):
LGTM.

wmmc88:
> I don't see any documentation updates in this PR. Could you confirm you've looked over the README and other documentation to determine if we need to update anything? For docs, we might need a follow-on PR to [the Mbed CLI 2 docs in mbed-os-5-docs](https://github.com/ARMmbed/mbed-os-5-docs/tree/development/docs/tools/mbed_cli_2), given that the code and docs live in separate repos.

I believe this pr made documentation on [mbed cli 2](https://os.mbed.com/docs/mbed-os/v6.9/build-tools/use.html) incorrect. Specifically the build folder paths (previously just `cmake_build`) in step 2 and 3 of the `Build the project` heading.

src/mbed_tools/cli/configure.py
```diff
@@ -48,10 +48,11 @@ def configure(toolchain: str, mbed_target: str, program_path: str, mbed_os_path:
         program_path: the path to the local Mbed program
         mbed_os_path: the path to the local Mbed OS directory
     """
+    cmake_build_subdir = pathlib.Path(mbed_target.upper(), "develop", toolchain.upper())
```
wmmc88, line 51: From the PR description and commit msg, it says that configs and builds should now live under `cmake_build/TARGET/PROFILE/TOOLCHAIN`, but this line makes `mbed-tools configure` always generate the `mbed_config.cmake` under `cmake_build/TARGET/develop/TOOLCHAIN`. Why is this?

PR #190, opened 2021-02-15
project: Add support for SSH syntax in import

### LINKED ISSUE #169: mbed-tools import fails to import an example with ssh url
### Description

mbed-tools version: **5.0.0**

Command:

`mbed-tools -vv import git@github.com:ARMmbed/mbed-os-example-blinky.git`

Expected:

mbed-os-example-blinky example cloned onto a local machine.

 

Output:

```

Cloning Mbed program 'git@github.com:ARMmbed/mbed-os-example-blinky.git'
Resolving program library dependencies.
ERROR: Cloning git repository from url 'https://github.com/armmbed/git@github.com:ARMmbed/mbed-os-example-blinky.git' failed. Error from VCS: Cmd('git') failed due to: exit code(128)
cmdline: git clone --progress -v https://github.com/armmbed/git@github.com:ARMmbed/mbed-os-example-blinky.git mbed-os-example-blinky.git

More information may be available by using the command line option '-vvv'.

```

### Issue request type

- [ ] Enhancement
- [x] Bug

wernerlewis:
### Description

The parser in the `import` subcommand causes errors when using scp-like `user@host:directory` SSH syntax, as it is not recognized by `urllib.parse` as a standard URL, and is treated as a project name instead. This adds a check for non-standard URLs, using the URL if it meets git scp-like requirements. This does not affect importing ARMmbed projects by name.

Fixes #169

### Test Coverage

### DIFF
--- a/src/mbed_tools/project/mbed_program.py
+++ b/src/mbed_tools/project/mbed_program.py
@@ -113,6 +113,9 @@ def parse_url(name_or_url: str) -> Dict[str, str]:
     url_obj = urlparse(name_or_url)
     if url_obj.hostname:
         url = url_obj.geturl()
+    elif ":" in name_or_url.split("/", maxsplit=1)[0]:
+        # If non-standard and no slashes before first colon, git will recognize as scp ssh syntax
+        url = name_or_url
     else:
         url = f"https://github.com/armmbed/{url_obj.path}"
     # We need to create a valid directory name from the url path section.
--- a/news/169.bugfix
+++ b/news/169.bugfix
@@ -1,0 +1,1 @@
+Support use of user@host:directory syntax with the import subcommand.
--- a/tests/project/test_mbed_program.py
+++ b/tests/project/test_mbed_program.py
@@ -127,6 +127,12 @@ class TestParseURL(TestCase):
         self.assertEqual(data["url"], url)
         self.assertEqual(data["dst_path"], "mbed-os-example-numskull")
 
+    def test_creates_valid_dst_dir_from_ssh_url(self):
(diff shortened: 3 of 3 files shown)


src/mbed_tools/project/mbed_program.py
```diff
@@ -113,6 +113,9 @@ def parse_url(name_or_url: str) -> Dict[str, str]:
     url_obj = urlparse(name_or_url)
     if url_obj.hostname:
         url = url_obj.geturl()
+    elif ":" in name_or_url.split("/", maxsplit=1)[0]:
```
rwalton-arm, line 116: Does this cover all valid `git` urls? Should we try and use a library to parse the url instead? There are at least 2 libraries that claim to handle parsing `git` urls on pypi, although I have no idea about their quality.
https://github.com/git/git/blob/master/t/t0110-urlmatch-normalization.sh here are the test cases `git` uses for its url parser. I don't know how many of these formats we need to support or if your changes mean we now support all of them.

wernerlewis: From `git clone` docs, for URLs `git` supports either `scheme://` formats, which `urlparse` will parse, or the `[user@]host:` format, which is only recognized by `git` if there is a no slash before the first colon. Adding the check means that format will now be supported in addition to the cases `urlparse` covers. I did look at using an available parsing library, but it might be unnecessary if we're just passing the possible URLs to `git`.

rwalton-arm: OK, seems fine then. Thanks for explaining.

rwalton-arm (approved):

PR #192, opened 2021-02-16
config: Check namespace when applying overrides

Patater:
### Description

Previously, when applying overrides, we were finding the first config
setting with a name that matches the the override name. When different
setting namespaces contained settings with the same name, this led to
applying overrides to only the first setting rather than the correct
setting. To fix this, check both name and namespace when searching for
settings to override.

### Test Coverage

- [X]  This change is covered by existing or additional automated tests.

### DIFF
--- a/news/191.bugfix
+++ b/news/191.bugfix
@@ -1,0 +1,1 @@
+Take namespaces into account when applying target overrides.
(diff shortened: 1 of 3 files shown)


rwalton-arm (approved):
LGTM

PR #198, opened 2021-02-19
devices: Add IDs for devices when multiple of the same type are connected

### LINKED ISSUE #181: mbedtools detect with multiple targets
### Description

What is the expected behavior when same target is connected several times?

"old" tool:
```
$ mbedls
| platform_name | platform_name_unique | mount_point | serial_port | target_id                | interface_version |
|---------------|----------------------|-------------|-------------|--------------------------|-------------------|
| NUCLEO_WB55RG | NUCLEO_WB55RG[0]     | E:          | COM26       | 08390221104868612F0CF165 | 0221              |
| NUCLEO_WB55RG | NUCLEO_WB55RG[1]     | F:          | COM42       | 08390221104968152F0DF111 | V2J37M26          |
```
=> platform_name_unique could be used

Now:
```
$ mbedtools detect
Board name     Serial number             Serial port    Mount point(s)    Build target(s)
-------------  ------------------------  -------------  ----------------  -----------------
NUCLEO-WB55RG  066dff535456807867075351  COM42          F:                NUCLEO_WB55RG
NUCLEO-WB55RG  066dff545057717867195937  COM26          E:                NUCLEO_WB55RG
```

@MarceloSalazar 

### Issue request type

- [x] Enhancement
- [ ] Bug

wernerlewis:
### Description

Refactor device detection to allow for use of multiple of the same device, without having to manually remove, and to clarify default flash behavior in this case.

Previously, if a user had multiple of the same `mbed-target` connected, the behavior of `mbed-tools` was to use the first device found which matches the given `mbed_target`, regardless of how many are connected. That leads to unclear behavior with `compile -f/--sterm`, and `sterm` subcommands, and requires manual disconnecting of devices to flash or start a serial terminal with a specific device.

Modified the default behavior of `compile -f` to find all connected devices of the target type, and flash all that are found, rather than just flashing the first one found.

Allow users to specify a device to use with an ID in `compile` and `sterm`: `mbed_target[ID]`. This allows use of `-f/--sterm` with any connected device even if others of the same type are connected. The unique build targets with the IDs are displayed with `detect`.

For example the `detect` output from #181 would now be:
```
Board name     Serial number             Serial port    Mount point(s)    Build target(s)
-------------  ------------------------  -------------  ----------------  -----------------
NUCLEO-WB55RG  066dff535456807867075351  COM42          F:                NUCLEO_WB55RG[0]
NUCLEO-WB55RG  066dff545057717867195937  COM26          E:                NUCLEO_WB55RG[1]
```

Added a print statement to `compile -f` to tell the user how many files were copied, and whether they were `.BIN` or `.HEX`.

Fixes #181

### Test Coverage

### DIFF
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -7,6 +7,8 @@ import os
 import pathlib
 import shutil
 
+from typing import Optional, Tuple
+
 import click
 
 from mbed_tools.build import build_project, generate_build_system, generate_config, flash_binary
@@ -84,6 +86,8 @@ def build(
        baudrate: Change the serial baud rate (ignored unless --sterm is also given).
     """
     _validate_target_and_toolchain_args(mbed_target, toolchain)
+    mbed_target, target_id = _get_target_id(mbed_target)
+
     cmake_build_subdir = pathlib.Path(mbed_target.upper(), profile.lower(), toolchain.upper())
     if mbed_os_path is None:
         program = MbedProgram.from_existing(pathlib.Path(program_path), cmake_build_subdir)
@@ -103,8 +107,8 @@ def build(
     build_project(build_tree)
 
     if flash or sterm:
-        if sterm:
-            devices = [find_connected_device(mbed_target)]
+        if target_id is not None or sterm:
+            devices = [find_connected_device(mbed_target, target_id)]
         else:
             devices = find_all_connected_devices(mbed_target)
 
@@ -131,3 +135,12 @@ def _validate_target_and_toolchain_args(target: str, toolchain: str) -> None:
         raise click.UsageError(
             "Both --toolchain and --mbed-target arguments are required when using the compile subcommand."
         )
+
+
+def _get_target_id(target: str) -> Tuple[str, Optional[int]]:
+    if "[" in target:
+        target_name, target_id = target.replace("]", "").split("[", maxsplit=1)
+        if target_id.isdigit() and int(target_id) >= 0:
+            return (target_name, int(target_id))
+        raise click.ClickException("When using the format mbed-target[ID], ID must be a positive integer or 0.")
+    return (target, None)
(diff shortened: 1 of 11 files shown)


src/mbed_tools/cli/build.py
```diff
@@ -111,6 +111,7 @@ def build(
     if flash:
         for dev in devices:
             flash_binary(dev.mount_points[0].resolve(), program.root, build_tree, mbed_target, hex_file)
+        click.echo(f"{len(devices)} file(s) copied ({'.HEX' if hex_file else '.BIN'}).")
```
rwalton-arm, line 122: Can we not just display the full path(s) of the files copied? This log message looks like it will display e.g "1 file(s) copied (.HEX)", which is more useful than no message but isn't really that useful for debugging etc.

src/mbed_tools/cli/list_connected_devices.py
```diff
@@ -49,25 +49,40 @@ def _sort_devices(devices: Iterable[Device]) -> Iterable[Device]:
     return sorted(devices, key=attrgetter("mbed_board.board_name", "serial_number"))
 
 
+def _get_ids_devices(devices: Iterable[Device]) -> Iterable[Tuple[Optional[int], Device]]:
```
rwalton-arm, line 52: Isn't `_get_devices_ids` a more natural way to name this?

rwalton-arm, line 52: We should be more specific about the return type, if it's a list we should specify that. An `Iterable` could be list, generator, string, etc.

src/mbed_tools/cli/list_connected_devices.py
```diff
@@ -80,12 +95,17 @@ def _build_json_output(devices: Iterable[Device]) -> str:
...
+                    "build_targets": _get_build_targets(board, id),
                 },
             }
         )
     return json.dumps(devices_data, indent=4)
 
 
-def _get_build_targets(board: Board) -> Iterable[str]:
-    return [f"{board.board_type}_{variant}" for variant in board.build_variant] + [board.board_type]
+def _get_build_targets(board: Board, identifier: Optional[int]) -> Iterable[str]:
```
rwalton-arm, line 105: More specific return type please.

rwalton-arm (changes requested):

rwalton-arm (approved):

PR #207, opened 2021-03-02
config: Pass mbed-os-path to generate_config

### LINKED ISSUE #204: Mbed OS path not searched when processing config
### Description
When we added the `--mbed-os-path` option to `configure` and `compile` we didn't add it to the list of search paths for the config processing tool. Currently, the tool always searches for config files from the project root down, and `mbed-os-path` could be outside of the project root.

This causes issues with Greentea tests, as the Greentea test "project root" is a subdirectory of `mbed-os`, so the tool fails to find any `mbed_lib.json` files as it's only searching from the "project root" down.

We should pass this `mbed-os-path` argument to the config processing tool so it can search mbed-os for config files even when mbed-os is outside of the "project root".

### Issue request type

- [ ] Enhancement
- [x] Bug

rwalton-arm:
Previously the mbed-os-path wasn't being passed to assemble_config.
This meant if mbed-os was in a directory above the project root no
mbed_lib.json files would be detected as assemble_config was always
searching for mbed_lib.json files from the project root down.

This commit modifies assemble_config to take an iterable of search
paths it will scan for config files.

Fixes #204

### Test Coverage

### DIFF
--- a/src/mbed_tools/build/_internal/config/assemble_build_config.py
+++ b/src/mbed_tools/build/_internal/config/assemble_build_config.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """Configuration assembly algorithm."""
+import itertools
+
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterable, List, Optional, Set
@@ -12,7 +14,7 @@ from mbed_tools.build._internal.config import source
 from mbed_tools.build._internal.find_files import LabelFilter, RequiresFilter, filter_files, find_files
 
 
-def assemble_config(target_attributes: dict, mbed_program_directory: Path, mbed_app_file: Optional[Path]) -> Config:
+def assemble_config(target_attributes: dict, search_paths: Iterable[Path], mbed_app_file: Optional[Path]) -> Config:
     """Assemble config for given target and program directory.
 
     Mbed library and application specific config parameters are parsed from mbed_lib.json and mbed_app.json files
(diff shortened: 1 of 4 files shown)


src/mbed_tools/build/_internal/config/assemble_build_config.py
```diff
@@ -24,8 +26,15 @@ def assemble_config(target_attributes: dict, mbed_program_directory: Path, mbed_
...
     Then we parse all the required mbed_lib config and finally apply the app overrides.
+
+    Args:
+        target_attributes: Mapping of target specific config parameters.
+        search_paths: Iterable of paths to search for mbed_lib.json files.
+        mbed_app_file: The path to mbed_app.json. This can be None.
     """
-    mbed_lib_files = find_files("mbed_lib.json", mbed_program_directory)
+    mbed_lib_files = list(
+        set(itertools.chain.from_iterable(find_files("mbed_lib.json", path) for path in search_paths))
```
Patater, line 36: What happens if multiple lib files are found? Using `set` would mean the order isn't guaranteed, so sometimes it might work and sometimes maybe not. I'd guess the user expectation is to use the deepest level mbed-os found. So, if the directory structure is `mbed-os/examples/some-example/mbed-os`, `mbed-os/examples/some-example/mbed-os` would be used and the top-level `mbed-os` would not be used.

rwalton-arm: I'm not sure I understand what you mean. Multiple `mbed_lib.json` files are nearly always found. The intermediate cast to `set` removes any duplicate `mbed_lib.json` paths. It shouldn't affect how deeply we traverse into the directory tree. Maybe I'm misunderstanding the issue?

Anyway, the reasoning for the deduplication is: now that we always search `mbed_os.root` and `project.root` for all `mbed_lib.json` files, if `mbed_os.root` was a sub-directory of `project.root` we would search the `mbed-os` tree twice. Obviously this would cause many duplicate `mbed_lib.json` paths to end up in `mbed_lib_files`. If we had a directory structure like your example, I would expect the `mbed_lib.json` paths in the "top level" `mbed-os` to compare `neq` to the paths in the deepest level `mbed-os` so they would all be collected if both `mbed-os` folders were in either the `project.root` tree or the `mbed_os.root` tree.

Patater: Sorry, I meant `mbed-os.lib`, which this line doesn't have anything to do with.

Patater (approved):

PR #218, opened 2021-03-05
tests: Add cases for board database requests

wernerlewis:
### Description

To make sure we're supporting proxies, test cases are added to check that `http_proxy` and `https_proxy` will be read from when accessing the online board database.

Tests are also added to improve coverage with the board database request code, to make sure that the correct warnings are being logged.

### Test Coverage

### DIFF
--- a/tests/targets/_internal/test_board_database.py
+++ b/tests/targets/_internal/test_board_database.py
@@ -4,71 +4,75 @@
 #
 """Tests for `mbed_tools.targets._internal.board_database`."""
 
-from unittest import TestCase, mock
+from unittest import mock
 
-import requests_mock
+import logging
+import pytest
 
-# Unit under test
 import mbed_tools.targets._internal.board_database as board_database
 from mbed_tools.targets.env import env
 
 
-class TestGetOnlineBoardData(TestCase):
+class TestGetOnlineBoardData:
     """Tests for the method `board_database.get_online_board_data`."""
 
-    @requests_mock.mock()
-    @mock.patch("mbed_tools.targets._internal.board_database.logger.warning", autospec=True)
-    @mock.patch("mbed_tools.targets._internal.board_database.logger.debug", autospec=True)
-    def test_401(self, mock_request, logger_debug, logger_warning):
+    def test_401(self, caplog, requests_mock):
         """Given a 401 error code, BoardAPIError is raised."""
-        mock_request.get(board_database._BOARD_API, status_code=401, text="Who are you?")
-        with self.assertRaises(board_database.BoardAPIError):
+        caplog.set_level(logging.DEBUG)
+        requests_mock.get(board_database._BOARD_API, status_code=401, text="Who are you?")
+        with pytest.raises(board_database.BoardAPIError):
             board_database.get_online_board_data()
-        self.assertTrue("MBED_API_AUTH_TOKEN" in str(logger_warning.call_args), "Auth token should be mentioned")
-        self.assertTrue("Who are you?" in str(logger_debug.call_args), "Message content should be in the debug message")
+        assert any(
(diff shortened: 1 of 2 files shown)


tests/targets/_internal/test_board_database.py
```diff
@@ -85,11 +85,19 @@ def test_no_auth_header_set_with_empty_token_var(self, requests):
...
+        logger_warning.assert_not_called()
+
+    @mock.patch("mbed_tools.targets._internal.board_database.logger.warning", autospec=True)
     @mock.patch("mbed_tools.targets._internal.board_database.requests.get")
-    def test_raises_tools_error_on_connection_error(self, get):
+    def test_raises_tools_error_on_connection_error(self, get, logger_warning):
         get.side_effect = board_database.requests.exceptions.ConnectionError
         with self.assertRaises(board_database.BoardAPIError):
             board_database._get_request()
+        logger_warning.assert_called_once()
```
Patater, line 100: This'll work, but there is a [pytest `caplog` builtin fixture](https://docs.pytest.org/en/stable/logging.html#caplog-fixture) you could use for more elegance.

tests/targets/_internal/test_board_database.py
```diff
@@ -115,6 +115,22 @@ def test_logs_error_on_requests_proxy_error(self, get, logger_warning):
...
+        self.assertEqual(
+            mock_request.last_request.proxies, {"http": "http://proxy:8080", "https": "https://proxy:8080"}
+        )
+
+    @mock.patch.dict("os.environ", {"http_proxy": "http://proxy:8080", "https_proxy": "https://proxy:8080"})
+    @mock.patch("mbed_tools.targets._internal.board_database.logger.warning", autospec=True)
+    def test_raises_proxy_error_with_invalid_proxy(self, logger_warning):
+        with self.assertRaises(board_database.BoardAPIError):
+            board_database._get_request()
+        self.assertTrue("connect to proxy" in str(logger_warning.call_args_list))
```
Patater, line 132: If you use `caplog`, `assert "connect to proxy" in caplog.text`

tests/targets/_internal/test_board_database.py
```diff
@@ -16,52 +18,65 @@
 class TestGetOnlineBoardData(TestCase):
     """Tests for the method `board_database.get_online_board_data`."""
 
+    @pytest.fixture(autouse=True)
+    def add_caplog(self, caplog):
+        self._caplog = caplog
```
Patater, line 23: I'm not sure this is right. It's not necessary to make our own caplog fixture, is it?

See how we use caplog at https://github.com/ARMmbed/mbed-tools/blob/master/tests/build/_internal/config/test_config.py#L114-L119 for instance. There is no `add_caplog()` or `self._caplog` there, but we are able to use the pre-made pytest built-in `caplog`.

wernerlewis: It's because of it being used with `unittest.TestCase` methods, which can't directly receive the pytest fixture as an argument. I could refactor it so that they're not `TestCase` methods, which would remove the need for `self._caplog`.

Patater: We've been moving away from unittest in most of the other modules. I'd recommend we refactor to just use pytest directly as an early commit in this PR.

tests/targets/_internal/test_board_database.py
```diff
@@ -4,71 +4,76 @@
...
 """Tests for `mbed_tools.targets._internal.board_database`."""
 
-from unittest import TestCase, mock
+from unittest import mock
 
-import requests_mock
+import logging
+import pytest
 
 # Unit under test
```
rwalton-arm, line 12: Pre-existing: this comment is silly. It would be good if we deleted it.

PR #219, opened 2021-03-08
ci: Fix preprelease when no docs are generated

rwalton-arm:
### Description

We do not document _internal packages. The prep-release script expected
we would always generate new API documentation when we made a release.

Make the prep-release script a bit more intelligent about when it will
create and amend commits during the release process.

### Test Coverage


ci_scripts/prep-release
```diff
@@ -2,36 +2,54 @@
...
-# Towncrier will fail if it detects a news file that has not been committed to
-# the repository. So we need to ensure the news file generated by
-# sync_board_database.py was committed before running towncrier.
-git commit -m "Prepare mbed-tools release ${NEW_VERSION}"
+# Update the index so we can check any recently staged changes
+git update-index -q --ignore-submodules --refresh
+
+if ! git diff-index --cached --quiet HEAD --ignore-submodules --
+then
+    printf "Found staged changes to commit after docs and target database update: $(git diff-index --cached --name-status -r --ignore-submodules HEAD)\n"
```
Patater, line 28: What does `-r` do? I don't see it in the man page. Is it `--relative` or `--raw` or something else? I'm looking to understand the difference between what we print here and what is in our `if` above, so ensure we print what we are checking (other than purely cosmetic options).

ci_scripts/prep-release
```diff
@@ -2,36 +2,54 @@
...
+fi
+
+printf "Tagging version ${NEW_VERSION}"
 git tag "${NEW_VERSION}"
 
 # Push the commit and tag to the remote. Relies on `GIT_TOKEN` environment
 # variable being set.
+printf "Pushing tag and master branch to remote\n"
 git push "https://${GIT_TOKEN}@github.com/ARMmbed/mbed-tools.git" master
 git push "https://${GIT_TOKEN}@github.com/ARMmbed/mbed-tools.git" --tags
```
Patater, line 55: Pre-existing issue: this pushes all tags. We only need to push `"${NEW_VERSION}"`. I'd recommend `git push "https://${GIT_TOKEN}@github.com/ARMmbed/mbed-tools.git"
 "${NEW_VERSION}"`

ci_scripts/prep-release
rwalton-arm: `-r` should be `--raw`, but `--raw` seems to be default now (and they don't document `-r` anymore). I'll remove this as it's not needed.

Patater (approved):

PR #232, opened 2021-03-12
tests: Remove redundant board db proxy test

wernerlewis:
### Description

When adding tests for proxy support, a test was added which uses `requests` with a non-existent proxy. This is effectively a test of `requests`, rather than a unit test of `mbed-tools`, and is redundant as other tests cover this proxy error using mocks.

### Test Coverage

### DIFF
--- a/news/20210312114544.misc
+++ b/news/20210312114544.misc
@@ -1,0 +1,1 @@
+Remove redundant test of proxy support in test_board_database.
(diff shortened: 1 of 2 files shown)


Patater (approved):
LGTM

Cool to see the coverage number stay the same. That helps give us confidence this test was redundant.

Issue #216, opened 2021-03-04
Support `OUTPUT_EXT` from targets.json


LDong-Arm:
### Description

Mbed OS's `targets.json` contains a parameter [`OUTPUT_EXT`](https://github.com/ARMmbed/mbed-os/blob/76b2902f48197f95a0c7f5e19b748cc9dbfb052e/targets/targets.json#L3103) - if unspecified, Mbed CLI 1 generates a `.bin` image, otherwise it generates the specified format (usual `.hex`). As this is not passed down CMake by mbed-tools, we currently [generate both `.bin` and `.hex`](https://github.com/ARMmbed/mbed-os/blob/76b2902f48197f95a0c7f5e19b748cc9dbfb052e/CMakeLists.txt#L147-L161).

For some targets, this can be quite misleading to users:
* Post-build hooks (e.g. bootloader, signing, etc.) only run on one of `.bin` or `.hex`, not both. This leaves only one of the two images usable. But because both have the same base name, users may have the impression that they can use either.
* Some targets only support one format (e.g. the NRF52840 MCU only supports `.hex`). Btw, I guess the `-f` option of `mbed-tools` would also need this to determine which image to flash?

If mbed-tools can pass `OUTPUT_EXT` (if set) to the config file, then we can rework the CMake part to handle this.

@ARMmbed/mbed-b-tools @Patater @0xc0170 @hugueskamba

### Issue request type

- [x] Enhancement
- [ ] Bug

LDong-Arm:
PRs created:
* https://github.com/ARMmbed/mbed-tools/pull/242
* https://github.com/ARMmbed/mbed-os/pull/14472

PR #242, opened 2021-03-25
config: Add `OUTPUT_EXT`

LDong-Arm:
### Description

Fixes: #216 

The optional `OUTPUT_EXT` field in Mbed OS `targets.json` specifies the format of the output image. This is important because
* some targets only support one image format
* each post-binary hook outputs one image only

This avoids confusion of having both .hex and .bin images generated while only one of them is usable.

Note: The CMake change in Mbed OS is https://github.com/ARMmbed/mbed-os/pull/14472

### Test Coverage

### DIFF
--- a/news/216.bugfix
+++ b/news/216.bugfix
@@ -1,0 +1,1 @@
+Add a CMake variable `MBED_OUTPUT_EXT` for the output image extension.
--- a/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
+++ b/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
@@ -9,6 +9,7 @@ set(MBED_TARGET "{{target_name}}" CACHE STRING "")
 set(MBED_CPU_CORE "{{core}}" CACHE STRING "")
 set(MBED_C_LIB "{{c_lib}}" CACHE STRING "")
 set(MBED_PRINTF_LIB "{{printf_lib}}" CACHE STRING "")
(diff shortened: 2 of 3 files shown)


LDong-Arm:
^^^ Is the coverage report expected? I did update `tests/build/test_generate_config.py`.

rwalton-arm:
> ^^^ Is the coverage report expected? I did update `tests/build/test_generate_config.py`.

Looks correct to me. The coverage tool just measures how many lines of Python code were hit. This commit doesn't change any Python code, it just edits the `jinja2` template. At the Python level we pass the entire config dictionary as the template context. So, the tool doesn't think the line coverage has changed, even though the extra test case was added. Another example of why 100% line coverage != 100% state coverage.

news/20210323151402.misc
```diff
@@ -1 +1,2 @@
 Update form factor ARDUINO to ARDUINO_UNO in the custom targets travis test as ARDUINO has been deprecated.
+Add a CMake variable for the output image extension
```
rwalton-arm, line 2: Can we add another, separate news fragment for this change? That way `towncrier` will format the `CHANGELOG.md` correctly when we make a release. We also want a `.bugfix` extension instead of `.misc` for this change, so the release pipeline bumps the version correctly.

rwalton-arm (changes requested):

news/20210323151402.misc
LDong-Arm: Thanks, done

rwalton-arm (approved):
LGTM

PR #246, opened 2021-03-30
setup: Relax Click version requirement

### LINKED ISSUE #245: Remove pinning of `Click` version in `setup.py`
**Describe the bug**
We pin the `Click` version to 7.1 in `setup.py`. We should relax this requirement to `Click>=7.1` to try and prevent conflicts with other packages that could also specify a `Click` version. 

We rely on `Click` features which arrived in 7.1 so we do need at least this version, but later versions should also work.

rwalton-arm:
### Description

We were pinning the version of Click we depend on to 7.1. Pinning to a
specific version of the package is not ideal, as we would have to keep
updating the pinned version to avoid dependency conflicts with newer
Python packages that depend on later versions of Click. We depend on
some Click features that appeared in 7.1, so we do need at least this
version. Later versions of the package should also work for us, so we
don't need to specify the requirement so precisely.

This commit relaxes the requirement to Click 7.1 or greater. This should
minimise the potential for dependency conflicts with other packages.

Fixes #245 

### Test Coverage

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "python-dotenv",
-        "Click==7.1",
+        "Click>=7.1,<8",
         "pdoc3",
         "GitPython",
         "tqdm",
--- a/news/245.bugfix
+++ b/news/245.bugfix
@@ -1,0 +1,1 @@
+Unpin Click version to minimise the potential for dependency conflicts with other packages.


setup.py
```diff
@@ -37,7 +37,7 @@
     include_package_data=True,
     install_requires=[
         "python-dotenv",
-        "Click==7.1",
+        "Click>=7.1",
```
Patater, line 40: Assuming Click follows semantic versioning, should we specify >=7.1,<8? This would prevent us from automatically pulling in a Click that the Click developers feel is not backwards compatible with 7.x.

Patater (approved):

PR #247, opened 2021-03-30
devices: Detect and extract data from J-Link

wernerlewis:
### Description

When connecting with J-Link, some files we rely on for identifying the board may not be present. This adds parsers for `Segger.html` and `Board.html` files, and extracts the type of J-Link in use and the slug from the manufacturer's board URL. This slug will not always correspond to the Mbed slug, so an additional function is added to search for the matching board. This attempts to find a board in the database which matches any of `slug`, `board_name` or `board_type` with the extracted slug.

### Test Coverage

### DIFF
--- a/tests/devices/_internal/test_file_parser.py
+++ b/tests/devices/_internal/test_file_parser.py
@@ -8,10 +8,7 @@ from unittest import mock
 
 import pytest
 
-from mbed_tools.devices._internal.file_parser import (
-    OnlineId,
-    read_device_files,
-)
+from mbed_tools.devices._internal.file_parser import OnlineId, read_device_files
 
 
 class TestReadDeviceFiles:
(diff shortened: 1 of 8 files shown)


tests/devices/_internal/test_file_parser.py
```diff
@@ -111,6 +123,43 @@ def test_none_if_not_found(self, tmp_path):
...
+        assert read_device_files([tmp_path]).interface_details == {"Version": j_link_type}
+
+    def test_jlink_interface_empty_if_not_found(self, tmp_path):
+        pathlib.Path(tmp_path, "Segger.html").write_text("<html></html>")
+        pathlib.Path(tmp_path, "Board.html").write_text(build_board_html())
+
+        assert read_device_files([tmp_path]).interface_details == {}
+
+
+class TestSlugParser:
```
rwalton-arm, line 153: Can we test this through the public interface to the module? I'd prefer we didn't test implementation details directly as it will make our code easier to change later.

tests/devices/_internal/test_file_parser.py
```diff
@@ -111,6 +123,43 @@ def test_none_if_not_found(self, tmp_path):
         assert read_device_files([tmp_path]).online_id is None
 
 
+class TestReadsJlinkData:
+    def test_reads_target_name(self, tmp_path):
+        board_name = "FRDM-K66F"
+        pathlib.Path(tmp_path, "Segger.html").write_text(build_segger_html())
+        file_contents = build_board_html(f"https://www.dummy.com/{board_name}")
+        pathlib.Path(tmp_path, "Board.html").write_text(file_contents)
+        online_id = read_device_files([tmp_path]).online_id
```
rwalton-arm, line 132: Can we be consistent with the newline spacing? I usually write tests using the "Given, When, Then" style, leaving newlines between each section:
```
GIVEN: mocks, inputs, paths, etc

WHEN: run code under test

THEN: assert results are as expected
```

We do have a few preexisting inconsistencies, but I've been trying to tidy them up where I can. It is kind of a nit-pick but I find the tests a lot easier to read when we use line spacing to denote the different sections.

rwalton-arm (changes requested):

rwalton-arm (approved):
LGTM

PR #270, opened 2021-04-13
config: Support MBED_ROM_START and friends

### LINKED ISSUE #222: MBED_ROM_START and friends unavailable on Mbed CLI2
### Description

On Mbed CLI, the following symbols are generated and passed to compiler, linker, or both:

```sh
mbed compile -m NUMAKER_IOT_M487 -t ARM
```

**BUILD/NUMAKER_IOT_M487/ARM/.profile.c**:
```
{
    "flags": [
        ......
        "-DMBED_RAM_SIZE=0x28000",
        "-DMBED_RAM_START=0x20000000",
        "-DMBED_ROM_SIZE=0x80000",
        "-DMBED_ROM_START=0x0",
        ......
```

**BUILD/NUMAKER_IOT_M487/ARM/.profile.ld**:
```
{
    "flags": [
        ......
        "--predefine=\"-DMBED_BOOT_STACK_SIZE=1024\"",
        "--predefine=\"-DMBED_RAM_SIZE=0x28000\"",
        "--predefine=\"-DMBED_RAM_START=0x20000000\"",
        "--predefine=\"-DMBED_ROM_SIZE=0x80000\"",
        "--predefine=\"-DMBED_ROM_START=0x0\"",
        ......
```

But on Mbed CLI2, they are unavailable in `cmake_build/NUMAKER_IOT_M487/develop/ARM/mbed_config.cmake` or elsewhere.

```sh
mbed-tools compile -m NUMAKER_IOT_M487 -t ARM
```

### Issue request type

- [ ] Enhancement
- [x] Bug

### Mbed/Tool version

**mbed-os**: 6.8.0
**mbed-cli**: 1.10.5
**mbed-tools**:: 7.2.1

wernerlewis:
### Description

MBED_ROM_START/SIZE and MBED_RAM_START/SIZE may be included in config files, to define active memory regions. However, these are currently ignored in the config system, and so are not passed to the compiler.

These attributes are now extracted from the config files, and the definitions are added to `mbed_config.cmake`. Both SIZE and START are required together to define a region, and repeated definition of a memory region is ignored.

Fixes #222

### Test Coverage

### DIFF
--- a/src/mbed_tools/build/_internal/config/config.py
+++ b/src/mbed_tools/build/_internal/config/config.py
@@ -8,7 +8,7 @@ import logging
 from collections import UserDict
 from typing import Any, Iterable, Hashable, Callable, List
 
-from mbed_tools.build._internal.config.source import Override, ConfigSetting
+from mbed_tools.build._internal.config.source import Memory, Override, ConfigSetting
 
 logger = logging.getLogger(__name__)
 
@@ -18,13 +18,15 @@ class Config(UserDict):
 
     This object understands how to populate the different 'config sections' which all have different rules for how the
     settings are collected.
-    Applies overrides, appends macros and updates config settings.
+    Applies overrides, appends macros, updates memories, and updates config settings.
     """
 
     def __setitem__(self, key: Hashable, item: Any) -> None:
         """Set an item based on its key."""
         if key == CONFIG_SECTION:
             self._update_config_section(item)
+        elif key == MEMORIES_SECTION:
+            self._update_memories_section(item)
         elif key == OVERRIDES_SECTION:
             self._handle_overrides(item)
         elif key == MACROS_SECTION:
@@ -67,6 +69,20 @@ class Config(UserDict):
 
         self.data[CONFIG_SECTION] = self.data.get(CONFIG_SECTION, []) + config_settings
 
+    def _update_memories_section(self, memories: List[Memory]) -> None:
+        defined_memories = self.data.get(MEMORIES_SECTION, [])
+        for memory in memories:
+            logger.debug(f"Adding memory settings `{memory.name}: start={memory.start} size={memory.size}`")
+            prev_defined = next((mem for mem in defined_memories if mem.name == memory.name), None)
+            if prev_defined is None:
+                defined_memories.append(memory)
+            else:
+                logger.warning(
+                    f"You are attempting to redefine `{memory.name}` from {prev_defined.namespace}.\n"
+                    f"The values from `{memory.namespace}` will be ignored"
+                )
+        self.data[MEMORIES_SECTION] = defined_memories
+
     def _find_first_config_setting(self, predicate: Callable) -> Any:
         """Find first config setting based on `predicate`.
 
@@ -89,6 +105,7 @@ class Config(UserDict):
 
 CONFIG_SECTION = "config"
 MACROS_SECTION = "macros"
+MEMORIES_SECTION = "memories"
 OVERRIDES_SECTION = "overrides"
 
 
--- a/src/mbed_tools/build/_internal/config/source.py
+++ b/src/mbed_tools/build/_internal/config/source.py
@@ -28,8 +28,8 @@ def prepare(
 ) -> dict:
     """Prepare a config source for entry into the Config object.
 
-    Extracts config and override settings from the source. Flattens these nested dictionaries out into lists of
-    objects which are namespaced in the way the Mbed config system expects.
+    Extracts memory, config and override settings from the source. Flattens these nested dictionaries out into
+    lists of objects which are namespaced in the way the Mbed config system expects.
 
     Args:
         input_data: The raw config JSON object parsed from the config file.
@@ -45,6 +45,11 @@ def prepare(
     namespace = data.pop("name", source_name)
     for key in data:
         data[key] = _sanitise_value(data[key])
+
+    memories = _extract_memories(namespace, data)
+
+    if memories:
+        data["memories"] = memories
 
     if "config" in data:
         data["config"] = _extract_config_settings(namespace, data["config"])
@@ -76,6 +81,31 @@ class ConfigSetting:
     def __post_init__(self) -> None:
         """Convert the value to a set if applicable."""
         self.value = _sanitise_value(self.value)
+
+
+@dataclass
+class Memory:
+    """Representation of a defined RAM/ROM region."""
+
+    name: str
+    namespace: str
+    start: str
+    size: str
+
+    def __post_init__(self) -> None:
+        """Convert start and size to hex format strings."""
+        try:
+            self.start = hex(int(self.start, 0))
+        except ValueError:
+            raise ValueError(
+                f"Value of MBED_{self.name}_START in {self.namespace}, {self.start} is invalid: must be an integer"
+            )
+        try:
+            self.size = hex(int(self.size, 0))
+        except ValueError:
+            raise ValueError(
+                f"Value of MBED_{self.name}_SIZE in {self.namespace}, {self.size} is invalid: must be an integer"
+            )
 
 
 @dataclass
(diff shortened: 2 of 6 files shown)


jeromecoutant:
Hi

> These attributes are now extracted from the config files

Config files are targets/custom and mbed_app json files ?
Not tools/arm_pack_manager/index.json, right ?

Patater (approved):

tests/build/_internal/config/test_config.py
```diff
@@ -24,6 +25,53 @@ def test_raises_when_trying_to_add_duplicate_config_setting(self):
...
+        assert "values from `lib2` will be ignored" in caplog.text
+
+    def test_memory_attr_converted_as_hex(self):
+        input_dict = {"mbed_ram_size": "1024", "mbed_ram_start": "0x24000000"}
+
+        conf = Config(prepare(input_dict, source_name="lib"))
+
+        assert Memory("RAM", "lib", "0x24000000", "0x400") in conf["memories"]
+
+    def test_raises_memory_attr_not_integer(self):
```
rwalton-arm, line 57: The tests for `prepare` should probably be in `test_source.py`. This test suite is supposed to check the `Config` object behaves correctly. We only used `prepare` to format the input data for `Config` here previously,  now we're also testing `prepare` directly.

src/mbed_tools/build/_internal/config/source.py
```diff
@@ -128,6 +157,30 @@ def _extract_config_settings(namespace: str, config_data: dict) -> List[ConfigSe
...
+    memories = []
+    for mem in ["rom", "ram"]:
+        start_attr = f"mbed_{mem}_start"
+        size_attr = f"mbed_{mem}_size"
+        start = data.get(start_attr)
+        size = data.get(size_attr)
+
+        if size is not None and start is not None:
+            logger.debug(f"Extracting MBED_{mem.upper()} definitions in {namespace}: _START={start}, _SIZE={size}.")
+
```
rwalton-arm, line 170: The line spacing here seems a little off. We should probably remove this blank line.

src/mbed_tools/build/_internal/config/source.py
```diff
@@ -128,6 +157,30 @@ def _extract_config_settings(namespace: str, config_data: dict) -> List[ConfigSe
...
+            memory = Memory(mem.upper(), namespace, start, size)
+            memories.append(memory)
+            del data[start_attr]
```
rwalton-arm, line 173: I'm not sure why we remove the attributes from `data` here.

src/mbed_tools/build/_internal/config/config.py
```diff
@@ -67,6 +69,20 @@ def _update_config_section(self, config_settings: List[ConfigSetting]) -> None:
 
         self.data[CONFIG_SECTION] = self.data.get(CONFIG_SECTION, []) + config_settings
 
+    def _update_memories_section(self, memories: List[Memory]) -> None:
+        defined_memories = self.data.get(MEMORIES_SECTION, [])
+        for memory in memories:
+            logger.debug(f"Adding memory settings `{memory.name}: start={memory.start} size={memory.size}`")
+            prev_defined = next((mem for mem in defined_memories if mem.name == memory.name), None)
+            if prev_defined is None:
```
rwalton-arm, line 77: Can we reduce this to `if memory not in defined_memories` or is that not a strong enough check for this?

src/mbed_tools/build/_internal/config/source.py
```diff
@@ -78,6 +83,30 @@ def __post_init__(self) -> None:
...
+    namespace: str
+    start: str
+    size: str
+
+    def __post_init__(self) -> None:
+        """Convert start and size to hex format strings."""
+        self.start = self._as_hex(self.start, "START")
+        self.size = self._as_hex(self.size, "SIZE")
+
+    def _as_hex(self, value: str, attr: str) -> str:
```
rwalton-arm, line 100: We only use `self` for adding stuff to the logging message in this method. This seems like a weak reason to add another method to this class. I'd prefer a free-function here and possibly moving the error handling up into `__post_init__`

rwalton-arm (changes requested):

src/mbed_tools/build/_internal/config/config.py
rwalton-arm, line 77: To clarify, does this code not work for some reason?
```suggestion
        defined_memories = self.data.get(MEMORIES_SECTION, [])
        for memory in memories:
            logger.debug(f"Adding memory settings `{memory.name}: start={memory.start} size={memory.size}`")
            if memory not in defined_memories:
```
Maybe we'll need to tweak the logging message slightly, but using `in` to check for duplicates should be faster and more readable. If I'm missing something and this code won't work then we can go with the original version.

wernerlewis: The more complicated check is just to prevent multiple definitions for the same memory region, with a different start/size in the config. The `in` check would be faster, but leaves it to the preprocessor to warn about redefining MBED_ROM_START etc. It might not be a frequent enough case to warrant the speed/readability cost of the original version though.

LDong-Arm (changes requested):
> Config files are targets/custom and mbed_app json files ?
> Not tools/arm_pack_manager/index.json, right ?

This is the question raised by @jeromecoutant above.

This change only works if a target's `mbed_ram_start` and friends are directly defined in [`targets.json`](https://github.com/ARMmbed/mbed-os/blob/master/targets/targets.json), and there are only 15 such targets.

But the vast majority of targets (116 of them) only have a `device_name` in `targets.json`, and the actual memory regions are defined in [`tools/arm_pack_manager/index.json`](https://raw.githubusercontent.com/ARMmbed/mbed-os/master/tools/arm_pack_manager/index.json) (a huge file of 517706 lines) and looked up based on `device_name`. The old tool does such look up, to support all targets. This huge file was imported from CMSIS according to https://github.com/ARMmbed/mbed-os/pull/3067/commits/167ed2bc121c9d3668d33558d7339492f7596c96 and modified whenever a new target is added.

@Patater In the long run what should we do? Keeping and using that huge file, or trying migrate a small amount of useful info into `targets.json` and removing it?

Before such refactoring happens (or if it won't happen), we might need to parse `tools/arm_pack_manager/index.json` too.

LDong-Arm:
Another issue with `tools/arm_pack_manager/index.json` is, despite having memory regions specified, the memory addresses and sizes are also hardcoded in each target's linker script (`.sct` or `.ld`) in most cases. Confusions arise if they disagree. Not to mention some targets (e.g. ARM_MUSCA_S1) only have memory regions in linker scripts, not in `tools/arm_pack_manager/index.json` or `targets.json`.

rwalton-arm:
> Another issue with `tools/arm_pack_manager/index.json` is, despite having memory regions specified, the memory addresses and sizes are also hardcoded in each target's linker script (`.sct` or `.ld`) in most cases. Confusions arise if they disagree. Not to mention some targets (e.g. ARM_MUSCA_S1) only have memory regions in linker scripts, not in `tools/arm_pack_manager/index.json` or `targets.json`.

I would suggest the best course of action is to migrate the necessary information to targets.json, and ignore pack_manager/index.json, as it seems to be out of date or incorrect in a number of cases. If the memory regions aren't configurable for certain targets, because they're hardcoded in the linkerscripts, then we should probably not define the properties in targets.json for those targets. That would mean the tool would at least emit a warning if a user attempts to override the memory region for a target where that isn't possible, which may reduce the potential other "confusions" later on.

LDong-Arm:
> I would suggest the best course of action is to migrate the necessary information to targets.json, and ignore pack_manager/index.json, as it seems to be out of date or incorrect in a number of cases.

Agreed.

>  If the memory regions aren't configurable for certain targets, because they're hardcoded in the linkerscripts, then we should probably not define the properties in targets.json for those targets. That would mean the tool would at least emit a warning if a user attempts to override the memory region for a target where that isn't possible, which may reduce the potential other "confusions" later on.

Most targets' linker scripts have hardcoded values, but they can be overridden with macros ([example](https://github.com/ARMmbed/mbed-os/blob/9738b27c7df897c29e9769911d6794ba3e5b3f19/targets/TARGET_ARM_SSG/TARGET_MUSCA_S1/device/TOOLCHAIN_ARMC6/musca_ns.sct#L24-L38)). The old mbed-cli sets memory region macros from either `targets.json` or `tools/arm_pack_manager/index.json` anyway. Any clean-up/alignment would be a large amount of work on its own.

At minimum we need to migrate memory regions from `tools/arm_pack_manager/index.json` to `targets.json` as suggested above, I guess.

Patater:
I've raised https://github.com/ARMmbed/mbed-tools/issues/271 for deciding what to do about `tools/arm_pack_manager/index.json`. Please continue discussion there.

What do we need to move this PR forward, given we are limiting its scope to only memory map information defined in `targets.json`?

LDong-Arm:
> I've raised #271 for deciding what to do about `tools/arm_pack_manager/index.json`. Please continue discussion there.
> 
> What do we need to move this PR forward, given we are limiting its scope to only memory map information defined in `targets.json`?

Thanks. I will approve this PR as it's sufficient for the current scope. Maybe we can't close #222, because the target mentioned there is one that depends on #271?

LDong-Arm (approved):
LGTM. There's still a few comments from @rwalton-arm to be addressed?

src/mbed_tools/build/_internal/config/config.py
rwalton-arm: Thanks for explaining. Yes, I think the more complex check is warranted if it prevents us running into multiple definition warnings.

rwalton-arm (approved):
LGTM. We need to rebase then it should be good to merge.

PR #273, opened 2021-04-27
CMake: Render CMakeLists template from mbed-os

harmut01:
### Description

`CMakeLists.tmpl` has been copied into `tools/cmake` in mbed-os. This is
the copy that should be used when creating a project from new i.e.
`mbed-tools new`. The cached copy of the template should only be used
if fetching the new template fails.

### Test Coverage

### DIFF
--- a/src/mbed_tools/project/_internal/render_templates.py
+++ b/src/mbed_tools/project/_internal/render_templates.py
@@ -4,10 +4,14 @@
 #
 """Render jinja templates required by the project package."""
 import datetime
+import logging
 
 from pathlib import Path
 
 import jinja2
+import requests
+
+logger = logging.getLogger(__name__)
 
 TEMPLATES_DIRECTORY = Path("_internal", "templates")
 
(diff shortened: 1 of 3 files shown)


harmut01:
PR still requires tests to be added and a little rework. Will close and reopen once all the changes are ready for review.

src/mbed_tools/project/_internal/render_templates.py
```diff
@@ -19,11 +23,18 @@ def render_cmakelists_template(cmakelists_file: Path, program_name: str) -> None
...
     """
-    cmakelists_file.write_text(
-        render_jinja_template(
-            "CMakeLists.tmpl", {"program_name": program_name, "year": str(datetime.datetime.now().year)}
-        )
-    )
+    context = {"program_name": program_name, "year": str(datetime.datetime.now().year)}
+
+    try:
+        cmakelists_url = "https://raw.githubusercontent.com/ARMmbed/mbed-os/master/tools/cmake/CMakeLists.tmpl"
```
wernerlewis, line 29: I'm unsure of what the purpose is of having the template in both places if we're going to always try to use the current template. For example, a user uses mbed-tools 7.x which comes with a template that supports Mbed OS 6.y, this will try to fetch the template for the current version, 6.z, which may not be compatible. Even though they are using the tools which come with the correct template, they will have the wrong `CMakeLists` generated. But if they try and do it while offline, it will work.

Could instead get the template from the local `mbed-os` in use, which would require some other changes. We would also need a way to generate `CMakeLists` in a project with a shared copy of Mbed so the tools know where to look. For example, passing `--mbed-os-path` when using `new -c`.

The tools could fall back on a template which is not ideal, but backwards compatible, using `if(COMMAND ...)` to invoke required functions if they are defined in Mbed OS.

wernerlewis (changes requested):

Patater:
Closing in favor of not requiring CMake changes in all Mbed applications when Mbed OS updates.

PR #277, opened 2021-05-06
Refactor tests to use pytest fixtures and remove TestCase

wernerlewis:
### Description

In `test_mbed_program`, `test_project_data` and `test_render_templates`, temporary directories are created either with use of a decorator from `factories.py` or with repetitive code in each test case.

Use of `unittest.TestCase` is removed from these files, and the pytest `tmp_path` fixture is used to provide temporary directories.

### Test Coverage

### DIFF
--- a/news/20210512103446.misc
+++ b/news/20210512103446.misc
@@ -1,0 +1,1 @@
+Remove unittest.TestCase from tests and replace repetitive code with use of pytest fixtures.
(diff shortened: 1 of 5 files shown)


Patater:
First 3 commits look good. Given that we want a more stable CMake interface, to avoid requiring applications to change their CMakeLists.txt when upgrading Mbed OS, it doesn't make too much sense to make it easier to change the CMake interface.

Patater (approved):

PR #284, opened 2021-05-24
 Fix issue with memory region overrides

### LINKED ISSUE #283: mbed_rom_size values from mbed_app.json are ignored
**Describe the bug**

#270 has implemented mbed_rom_size support in targets.json file,
But it seems that if values is overwritten in the local mbed_app.json file, value is ignored

@wernerlewis 

**To Reproduce**

- choose a target with "mbed_rom_size" defined
- change the value in mbed_app.json: "target.mbed_rom_size"
- see in mbed_config.cmake

rwalton-arm:
### Description
MBED_ROM/RAM_START/SIZE overrides were not added to
mbed_config.cmake.

The reason mbed_app.json overrides were "ignored" is
because the overrides were applied to the already collected string
parameters and weren't applied to the extracted `Memory` objects we
expose in the template.

We shouldn't need special handling for mbed_rom/ram_start or
mbed_rom/ram_size as they are already collected (if present in
targets.json). So remove the special handling for memory regions.
A side effect of this approach is that we remove some of the validation
checks for START and SIZE both being defined. If we feel this is a 
necessary check we can add it back into the `Config` object.

Fixes #283

### Test Coverage

### DIFF
--- a/news/20210524113403.bugfix
+++ b/news/20210524113403.bugfix
@@ -1,0 +1,1 @@
+Fix issue with memory region overrides being ignored.
--- a/src/mbed_tools/build/_internal/cmake_file.py
+++ b/src/mbed_tools/build/_internal/cmake_file.py
@@ -5,6 +5,8 @@
 """Module in charge of CMake file generation."""
 import pathlib
 
+from typing import Any
+
 import jinja2
 
 from mbed_tools.build._internal.config.config import Config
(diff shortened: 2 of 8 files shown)


Patater (approved):
LGTM

Patater:
Travis passed, but failed to report back to GitHub.

PR #290, opened 2021-06-02
build: Automatically determine image format to flash

LDong-Arm:
### Description

The default image format is bin, but targets can override it with `"OUTPUT_EXT": "hex"` in `targets.json`. This is already taken care of in the CMake configuration generated by mbed-tools, but user still need to pass `--hex-image` if they want to run `mbed-tools compile` with `--flash` to flash a hex image.

This commit makes the image selection automatic based on `OUTPUT_EXT`.

Note: `--hex-file` is now redundant and thus removed. It does not make sense to have it anymore without also offering `--bin-file` for completeness.
    
The test cases for bin and hex image flashing have been updated accordingly. Coverage for `OUTPUT_EXT` has been improved, with both `mbed_config.cmake` and image format to flash checked.

### Test Coverage

- [x]  Manual testing has been performed (and evidence provided) as automated testing was not feasible.

The test case for hex image flashing has been updated accordingly.
Also manually tested on NRF52_DK (hex image) and DISCO_L475VG_IOT01A (bin image).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,4 +65,6 @@ name = "Misc"
 showcontent = false
 
 [tool.black]
+diff = true
+color = true
 line-length = 120
--- a/src/mbed_tools/build/config.py
+++ b/src/mbed_tools/build/config.py
@@ -5,20 +5,20 @@
 """Parses the Mbed configuration system and generates a CMake config script."""
 import pathlib
 
-from typing import Any
+from typing import Any, Tuple
 
 from mbed_tools.lib.json_helpers import decode_json_file
 from mbed_tools.project import MbedProgram
 from mbed_tools.targets import get_target_by_name
 from mbed_tools.build._internal.cmake_file import render_mbed_config_cmake_template
-from mbed_tools.build._internal.config.assemble_build_config import assemble_config
+from mbed_tools.build._internal.config.assemble_build_config import Config, assemble_config
 from mbed_tools.build._internal.write_files import write_file
 from mbed_tools.build.exceptions import MbedBuildError
 
 CMAKE_CONFIG_FILE = "mbed_config.cmake"
 
 
-def generate_config(target_name: str, toolchain: str, program: MbedProgram) -> pathlib.Path:
+def generate_config(target_name: str, toolchain: str, program: MbedProgram) -> Tuple[Config, pathlib.Path]:
     """Generate an Mbed config file at the program root by parsing the mbed config system.
 
     Args:
@@ -27,6 +27,7 @@ def generate_config(target_name: str, toolchain: str, program: MbedProgram) -> p
         program: The MbedProgram to configure.
 
     Returns:
+        Config object (UserDict)
         Path to the generated config file.
     """
     targets_data = _load_raw_targets_data(program)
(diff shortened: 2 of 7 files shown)


LDong-Arm:
@ARMmbed/mbed-os-core

LDong-Arm:
Now tests and code coverage should be good, the PR is ready for review.

pyproject.toml
```diff
@@ -65,4 +65,7 @@ name = "Misc"
 showcontent = false
 
 [tool.black]
+diff = true
+color = true
+check = true
```
rwalton-arm, line 70: I would remove `check = true` because that means `black` won't reformat my files when I run it locally. I want `black` to reformat my files when I run it locally.

src/mbed_tools/build/config.py
```diff
@@ -42,6 +42,21 @@ def generate_config(target_name: str, toolchain: str, program: MbedProgram) -> p
...
+    """Check if a target's image format is hex instead of bin.
+
+    Args:
+        target_name: Name of the target to check image format for.
+        program: The MbedProgram to extract targets data from.
+
+    Return:
+        True if the target's "OUTPUT_EXT" is "hex" in targets.json.
+    """
+    targets_data = _load_raw_targets_data(program)
```
rwalton-arm, line 55: Rather than loading and parsing `targets.json` again here just to find the output extension, should we refactor `generate_config` to separate the rendering and writing of the output file from the config assembly? Then we can just pull `OUTPUT_EXT` out of the `Config` object in `build`.

tests/build/test_generate_config.py
```diff
@@ -499,3 +497,44 @@ def test_config_parsed_when_mbed_os_outside_project_root(program_in_mbed_os_subd
...
+    assert 'MBED_OUTPUT_EXT "hex"' not in config_text
+    assert hex_image is False
+
+
+def test_output_ext_bin(program):
+    target_data = TARGET_DATA.copy()
+    target_data["OUTPUT_EXT"] = "bin"
+
+    program.files.custom_targets_json.write_text(json.dumps({"IMAGINARYBOARD": target_data}))
+    target = "IMAGINARYBOARD"
```
rwalton-arm, line 516: Should we swap these lines so you can dereference `target` in `program.files.custom_targets_json.write_text(json.dumps({"IMAGINARYBOARD": target_data}))`? This way you wouldn't have to write "IMAGINARYBOARD" twice.

rwalton-arm (changes requested):

src/mbed_tools/build/config.py
LDong-Arm: Returning a `Config` is a nice idea! I was wondering how to avoid loading json twice.

Currently generate_config returns the path to the output file, I'm changing it to also return the config itself.

LDong-Arm (commented):
@rwalton-arm Thanks, nice suggestions

rwalton-arm (approved):

rwalton-arm:
LGTM, just needs to be rebased.

LDong-Arm:
@rwalton-arm Thanks, rebased.

PR #292, opened 2021-06-22
configure/build: Support user-specified application configuration

### LINKED ISSUE #291: Missing `--app-config` option
**Describe the bug**
Mbed CLI 1 offers an `--app-config` option to let users specify the application configuration JSON instead of assuming `mbed_app.json`. This is not currently provided by mbed-tools, but useful when
* one application provides multiple configurations for different use cases
* different applications or tests share one common configuration (e.g. [experimental.json](https://github.com/ARMmbed/mbed-os/blob/master/TESTS/configs/experimental.json))

**To Reproduce**
Try to build an application (e.g. blinky) with `--app-config`, for example
```
mbed-tools compile -t GCC_ARM -m K64F --app-config mbed-os/TESTS/configs/experimental.json
```
and the error is
```
Error: no such option: --app-config
```

**Expected behavior**
An option `--app-config` is available to both `mbed-tools configure` and `mbed-tools compile`. The specified JSON is used for generating `mbed_config.cmake`.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Any
 - Version: Any

**Mbed (please complete the following information):**
 - Device: Any
 - Mbed OS Version: 6.11.0
 - Mbed CLI 2 Version: 7.23.0

**Additional context**
N/A

LDong-Arm:
### Description

Add an option `--app-config` to `configure` and `compile`, to allow users to specify an application configuration file. If unspecified, mbed-tools will use `mbed_app.json` if the latter exists.

Tests are provided.

Fixes #291

Also: Remove default arguments of `build()` that duplicate what we set with click.

### Test Coverage

### DIFF
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -41,6 +41,9 @@ from mbed_tools.sterm import terminal
     "--custom-targets-json", type=click.Path(), default=None, help="Path to custom_targets.json.",
 )
 @click.option(
+    "--app-config", type=click.Path(), default=None, help="Path to application configuration file.",
+)
+@click.option(
     "-f", "--flash", is_flag=True, default=False, help="Flash the binary onto a device",
 )
 @click.option(
@@ -63,6 +66,7 @@ def build(
     baudrate: int,
     mbed_os_path: str,
     custom_targets_json: str,
+    app_config: str,
 ) -> None:
     """Configure and build an Mbed project using CMake and Ninja.
 
@@ -76,6 +80,7 @@ def build(
        custom_targets_json: Path to custom_targets.json.
        toolchain: The toolchain to use for the build.
        mbed_target: The name of the Mbed target to build for.
+       app_config: the path to the application configuration file
        clean: Perform a clean build.
        flash: Flash the binary onto a device.
        sterm: Open a serial terminal to the connected target.
@@ -95,6 +100,8 @@ def build(
     click.echo("Configuring project and generating build system...")
     if custom_targets_json is not None:
         program.files.custom_targets_json = pathlib.Path(custom_targets_json)
+    if app_config is not None:
+        program.files.app_config_file = pathlib.Path(app_config)
     config, _ = generate_config(mbed_target.upper(), toolchain, program)
     generate_build_system(program.root, build_tree, profile)
 
--- a/src/mbed_tools/cli/build.py
+++ b/src/mbed_tools/cli/build.py
@@ -22,9 +22,10 @@ from mbed_tools.sterm import terminal
     "-t",
     "--toolchain",
     type=click.Choice(["ARM", "GCC_ARM"], case_sensitive=False),
+    required=True,
     help="The toolchain you are using to build your app.",
 )
-@click.option("-m", "--mbed-target", help="A build target for an Mbed-enabled device, e.g. K64F.")
+@click.option("-m", "--mbed-target", required=True, help="A build target for an Mbed-enabled device, e.g. K64F.")
 @click.option("-b", "--profile", default="develop", help="The build type (release, develop or debug).")
 @click.option("-c", "--clean", is_flag=True, default=False, help="Perform a clean build.")
 @click.option(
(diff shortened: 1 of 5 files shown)


src/mbed_tools/cli/build.py
```diff
@@ -124,13 +124,6 @@ def build(
         terminal.run(dev.serial_port, baudrate)
 
 
-def _validate_target_and_toolchain_args(target: str, toolchain: str) -> None:
-    if not all([toolchain, target]):
-        raise click.UsageError(
-            "Both --toolchain and --mbed-target arguments are required when using the compile subcommand."
-        )
-
-
```
LDong-Arm, line 133: The idea of removing this is to
* Take advantage of `required=True` of "click"
* Avoid having to set default arguments (to `""`) in `build()`

The slight drawback is "click" only checks one argument at a time. But eventually a user will need to pass both `--toolchain` and `--mbed-target` to not get any errors, so the outcome is the same.

This is also consistent with the `configure()` command.

LDong-Arm:
The Travis failure has nothing to do with this PR or mbed-tools. I've created a fix for it: https://github.com/ARMmbed/mbed-os/pull/14829

src/mbed_tools/cli/build.py
Patater: Please ensure this comment's info is in the commit message that makes this change. It's okay for commit messages to be a bit long.

tests/cli/test_build.py
```diff
@@ -116,18 +116,6 @@ def test_raises_if_gen_config_target_not_passed_when_required(
             self.assertIsNotNone(result.exception)
             self.assertRegex(result.output, "--mbed-target")
 
-    def test_raises_if_gen_config_target_toolchain_not_passed(
```
Patater, line 119: Does click not raise an exception if the target and toolchain aren't passed? What should we check for instead? What tests do we have that would make sure that if we remove "required=True" on accident, we still ensure each arg is required?

Patater (changes requested):

tests/cli/test_build.py
LDong-Arm: As per commit message:
```
Note: The test case where both target and toolchain options are
missing has been removed, because the "click" only checks one argument
at a time. The existing tests for one option missing (one for
toolchain, one for target) are sufficient.
```
I think two tests above the removed one already serve the purpose.

src/mbed_tools/cli/build.py
LDong-Arm: From the current commit message:
```
We set default values of arguments with "click", so there is no point
for `build()` to have default arguments again. Make `--toolchain`
and `--mbed-targets` required, to guarantee the variables exist.
```
Is this enough? Would you suggest also suggest mentiong the "drawback" and "consistency with configure()"?

tests/cli/test_build.py
LDong-Arm: So previously we had three tests
* target not passed
* toolchain not passed
* target & toolchain not passed

Because click checks only one argument at a time, only the first two make sense now. But they are still sufficient to ensure users pass both target & toolchain, to not have any problems.

Patater: OK, I didn't see we were testing them individually already and expected we had to split them.

src/mbed_tools/cli/build.py
Patater: Yes, drawbacks and consistency should be mentioned in the commit message, too. It helps gives better context around what the developer (you) was thinking at the time and what tradeoffs were made.

You felt you needed to comment in the PR beyond what was in the commit message. I am happy if you are satisfied the commit message containing all the information you wanted to share in the comment.

LDong-Arm: > You felt you needed to comment in the PR beyond what was in the commit message.

Good point, I think the tradeoff is worth mentioning in this case. Will update the commit message.

LDong-Arm:
Rebased and updated commit message

Patater (approved):

PR #301, opened 2021-07-20
config: Add `MBED_GREENTEA_TEST_RESET_TIMEOUT`

LDong-Arm:
### Description

Targets that take longer to boot (e.g. TF-M targets which boot MCUboot and TF-M before application) have `forced_reset_timeout` in mbed-os/targets/targets.json to increase timeout after reset before starting a Greentea test. This is previously handled by mbedgt, but as we move to CTest and mbedhtrun, a target's reset timeout needs to be available in CMake and passed to mbedhtrun.

### Test Coverage

### DIFF
--- a/news/20210720.feature
+++ b/news/20210720.feature
@@ -1,0 +1,1 @@
+Get `forced_reset_timeout` from mbed-os/targets/targets.json and add it to generated mbed_config.cmake as `MBED_GREENTEA_TEST_RESET_TIMEOUT` for use by CTest and mbedhtrun.
--- a/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
+++ b/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
@@ -10,6 +10,7 @@ set(MBED_CPU_CORE "{{core}}" CACHE STRING "")
 set(MBED_C_LIB "{{c_lib}}" CACHE STRING "")
 set(MBED_PRINTF_LIB "{{printf_lib}}" CACHE STRING "")
 set(MBED_OUTPUT_EXT "{{OUTPUT_EXT}}" CACHE STRING "")
+set(MBED_GREENTEA_TEST_RESET_TIMEOUT "{{forced_reset_timeout}}" CACHE STRING "")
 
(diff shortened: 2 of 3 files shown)


news/20210720144841.misc
```diff
@@ -0,0 +1 @@
+Get `forced_reset_timeout` from mbed-os/targets/targets.json and add it to generated mbed_config.cmake as `MBED_GREENTEA_TEST_RESET_TIMEOUT` for use by CTest and mbedhtrun.
```
Patater, line 1: This should be a `.feature`, not a `.misc`.

See guidance in our contributing guide: https://github.com/ARMmbed/mbed-tools/blob/master/CONTRIBUTING.md#news-files

Patater (changes requested):

news/20210720144841.misc
LDong-Arm: Done. As for the date in the filename, the example in the guide is precise to year, month and date only, not hours, minutes and seconds like in many existing PRs, so I shortened it to date too.

Patater (changes requested):
We should probably have a test for this in `tests/build/test_generate_config.py`.

news/20210720144841.misc
rwalton-arm: If we just use the date, and two PRs are raised on the same day that happen to add, for instance, a ".bugfix" newsfile, we'll get merge conflicts. That's why we usually use a more precise date in the filename, or even better, an issue number. The guide needs to be updated.

LDong-Arm: Fair enough. Probably precision to minute is more than enough - I often have to make up seconds 😅  because some other PRs have seconds.

LDong-Arm:
> We should probably have a test for this in `tests/build/test_generate_config.py`.

Done, tests added.

news/20210720144841.misc
rwalton-arm: I just use the date command e.g `nvim news/`date -u "+%Y%m%d%H%M%S"`.bugfix`

rwalton-arm (approved):

Patater (approved):

Patater (approved):

PR #303, opened 2021-07-29
Fix requires handling for baremetal config

rwalton-arm:
### Description

We were previously replacing the list of "requires" if we happened to
find a set of requires not in the app config. This meant occasionally
some config macros would not be defined for baremetal apps depending on
the order the config files were discovered and parsed.

Now we always append to the list of requires, which is how it's supposed
to work.

### Test Coverage

### DIFF
--- a/src/mbed_tools/build/_internal/config/assemble_build_config.py
+++ b/src/mbed_tools/build/_internal/config/assemble_build_config.py
@@ -84,13 +84,7 @@ def _get_app_filter_labels(mbed_app_data: dict, config: Config) -> None:
 
 
 def _get_file_filter_overrides(mbed_app_data: dict) -> dict:
-    return {
-        "overrides": [
-            override
-            for override in mbed_app_data.get("overrides", [])
-            if override.modifier or override.name == "requires"
-        ]
-    }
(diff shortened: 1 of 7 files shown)


src/mbed_tools/build/_internal/config/assemble_build_config.py
```diff
@@ -84,13 +84,7 @@ def _get_app_filter_labels(mbed_app_data: dict, config: Config) -> None:
...
 
 def _get_file_filter_overrides(mbed_app_data: dict) -> dict:
-    return {
-        "overrides": [
-            override
-            for override in mbed_app_data.get("overrides", [])
-            if override.modifier or override.name == "requires"
-        ]
-    }
+    return {"overrides": [override for override in mbed_app_data.get("overrides", []) if override.modifier]}
```
Patater, line 87: It may be worth mentioning in the commit message for "config: Remove support for target.requires" that this partially reverts 6a99b43bf11f8bebf52b6f3957a2653f8da03258

rwalton-arm: Mentioned.

Patater (approved):
LGTM

Issue #306, opened 2021-08-03
Add RAM1 configuration in mbed_config.tmpl


rajkan01:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
The existing "mbed_config.tmpl" is only generating MBED_RAM_START and MBED_RAM_SIZE and not generating MBED_RAM1_START and MBED_RAM1_SIZE, but most of the targets in targets.json has both RAM and RAM1 configuration.
This is a feature request to generate RAM1 configuration if the targets have a configuration in targets.json

PR #307, opened 2021-08-03
config: Add RAM1 memory region parameters to mbed_config.cmake

rajkan01:
### Description
Fixes: #306 
Add RAM1 memory region defines to mbed_config.tmpl as most of the targets
in targets.json has both RAM and RAM1 configurations

### Test Coverage

### DIFF
--- a/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
+++ b/src/mbed_tools/build/_internal/templates/mbed_config.tmpl
@@ -67,6 +67,12 @@ set(MBED_TARGET_DEFINITIONS{% for component in components %}
 {% if mbed_ram_size is defined %}
     MBED_RAM_SIZE={{ mbed_ram_size | to_hex }}
 {%- endif %}
+{% if mbed_ram1_start is defined %}
+    MBED_RAM1_START={{ mbed_ram1_start | to_hex }}
+{%- endif %}
+{% if mbed_ram1_size is defined %}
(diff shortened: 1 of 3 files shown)


src/mbed_tools/build/_internal/templates/mbed_config.tmpl
```diff
@@ -66,6 +66,12 @@ set(MBED_TARGET_DEFINITIONS{% for component in components %}
 {%- endif %}
 {% if mbed_ram_size is defined %}
     MBED_RAM_SIZE={{ mbed_ram_size | to_hex }}
+{%- endif %}
+{% if mbed_ram1_start is defined %}
+    MBED_RAM1_START={{ mbed_ram1_start | to_hex }}
```
rwalton-arm, line 71: Do we need to also support MBED_RAM_2..MBED_RAM_N? https://github.com/ARMmbed/mbed-os/blob/master/targets/TARGET_STM/TARGET_STM32L4/TARGET_STM32L443xC/TOOLCHAIN_ARM/stm32l443xc.sct#L50 seems to want MBED_RAM2_* at least.

LDong-Arm: @rwalton-arm Maybe it depends on whether we ever define `mbed_ram2_start` in targets.json?

Searching Mbed CLI 1 scripts (in mbed-os/tools/), it only has `mbed_ram_start`, not even ram1. So I wonder how (or if at all) `MBED_RAM1_START` is set by Mbed CLI 1.

PR #309, opened 2021-08-11
configure: Generate .mbedignore in build directory

### LINKED ISSUE #233: Users are building cli1 and cli2 within the same app
**Describe the bug**

I've received 2 questions with the similar topic within the last 24 hours - "my build is failing with a weird error" - a user first run cli2 and then later another day switched to cli1 and received an error.

A question is: should we fix this and avoid surprises if users use either cli1 or cli2 ? One suggestion that many of us came to was adding .mbedignore and add there `cmake_build` directory.  Is there better way to make this use case working without errors ?

**To Reproduce**

Build blinky with cli2 first, then build with cli1.

**Expected behavior**
Both builds would work as they work on the same codebase, in separate build directories. As we support both at the time being, we should allow users to build both or at least warn them if they are doing this (if we do not want to support the use case)

**Mbed (please complete the following information):**
mbed-tools, mbed-cli and Mbed OS - any version

LDong-Arm:
### Description

Mbed CLI 1 scans and builds all files of supported extensions in a project's directory and subdirectories, unless otherwise ignored. If a user has built a project once with mbed-tools and CMake then tries to build it with Mbed CLI 1, the CMake build directory gets accidentally picked up and results in errors.

To resolve this, let mbed-tools generate a .mbedignore file in the build directory.

Fixes #233.

### Test Coverage

### DIFF
--- a/news/233.bugfix
+++ b/news/233.bugfix
@@ -1,0 +1,1 @@
+Generate .mbedignore in a project's build directory, to prevent Mbed CLI 1 from picking up CMake build files.
--- a/src/mbed_tools/build/config.py
+++ b/src/mbed_tools/build/config.py
@@ -16,6 +16,7 @@ from mbed_tools.build._internal.write_files import write_file
 from mbed_tools.build.exceptions import MbedBuildError
 
 CMAKE_CONFIG_FILE = "mbed_config.cmake"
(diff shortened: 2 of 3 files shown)


rwalton-arm (approved):
LGTM

PR #322, opened 2022-01-10
Replace `licenseheaders` with `scancode`

rwalton-arm:
### Description

The `licenseheaders` tool has a number of issues which make it
unsuitable for our CI:

- It's not smart enough to know when it should apply a single year or a
  year range to touched files. No combination of options or templates
  can handle this.

- If a license header contains multiple contributor lines and year
  ranges, they are blindly overwritten with the license template. It's
  impossible to create a template to handle this.

- It can't be ran on all files because it will try to update all
  existing files whether they've actually been changed or not.

Replace `licenseheaders` with `scancode` which should be more
suitable. `scancode` is the tool Mbed OS uses on the main repo, so the
licensing checks should now be more aligned.

### Test Coverage


Patater:
Another benefit is if new languages are added, we don't need to explicitly turn on that support. For example, if we never had sh files before, and one were added, and we did not update the configuration files for license checking (say, we forgot), `scancode` would see and check the new sh file, but `licenseheaders` would not.

Patater (approved):

PR #327, opened 2022-07-20
config: Increase click package required version to 8.1

ATmobica:
### Description

The latest version (8.1.3) of the click package is supported in many
projects. E.g. in CHIP project where mbed-os port is implemented.
https://github.com/project-chip/connectedhomeip
In this case, we need to increase the limit of the click version to 8.1.

We need to also update the black package version to be
compatible with the click.

These changes required fix in TestNewCommand class. We need to use
`Path(<file>).resolve()` which is also use in latest click
implementation.
https://github.com/pallets/click/blob/dc918b48fb9006be683a684b42cc7496ad649b83/src/click/types.py#L853

### Test Coverage

- [x]  Additional tests are not required for this change (e.g. documentation update).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "python-dotenv",
-        "Click>=7.1,<8",
+        "Click>=7.1,<9",
         "GitPython",
         "tqdm",
         "tabulate",
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 19.10b0
+    rev: 22.6.0
     hooks:
       - id: black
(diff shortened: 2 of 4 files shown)


ATmobica:
@pan- link to this PR

pan- (approved):

setup.py
```diff
@@ -37,7 +37,7 @@
     include_package_data=True,
     install_requires=[
         "python-dotenv",
-        "Click>=7.1,<8",
+        "Click>=7.1,<=8.1",
```
Patater, line 40: Please use <9, as we expect any 8.x version to be OK. The current `<8` is meant to say "any future compatible version". If we are compatible with 8.1 now, with your changes, we should be good to go until 9.x.

Patater (changes requested):

Patater (approved):

